PR #4, opened 2016-08-30
Add CommandLineParsing framework

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2138

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Additional requirements for building and running from source:
 To run the standalone utility:
 
 ```sh
-$ java -jar fcrepo-import-export-<version>.jar
+$ java -jar fcrepo-import-export-<version>.jar --mode [import|export]
 ```
 
 To run the utility with Maven:
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ target/
 .idea/
 .checkstyle
 .idea/
+dependency-reduced-pom.xml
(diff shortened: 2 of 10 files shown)


ajs6f:
This isn't bad, but consider using [Airline](https://github.com/airlift/airline) for more concision and clarity.

README.md
```diff
@@ -15,7 +15,7 @@ Additional requirements for building and running from source:
 To run the standalone utility:
 
 ```sh
-$ java -jar fcrepo-import-export-<version>.jar
+$ java -jar fcrepo-import-export-<version>.jar --mode [import|export]
```
bseeger: Does java automatically convert `--mode` to the `-m` you're looking for below?

bseeger: Never mind, I see how it works.

awoods:
Re: Airline, I will not have the time right now to investigate redoing the CLI support... but if someone else is interested, great.
In a two week sprint, time is of the essence. I would suggest moving forward with this apache/commons-cli, and potentially revisit later.

escowles:
I agree with @awoods here — this looks perfectly functional, and I'd rather get something merged so we can move forward, and reconsider whether something like Airline would be better at a later date.

ajs6f:
I'll go so far as to file a ticket, but I'm certainly not going to do any work for it.

PR #6, opened 2016-08-30
Adding basic export implementation.

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2131

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,19 @@ Additional requirements for building and running from source:
 * Maven 3
 
 
-To run the standalone utility:
+To run the standalone export utility:
 
 ```sh
-$ java -jar fcrepo-import-export-<version>.jar --mode [import|export]
+$ java -jar fcrepo-export/target/fcrepo-export-<version>.jar --mode export --resource <URI> --binDir <dir> --descDir <dir>
 ```
+
+To control the format of the exported RDF, the extension and RDF language/serialization format can also be specified by adding, e.g.:
+
+```sh
+--rdfExt .jsonld --rdfLang application/ld+json
+```
+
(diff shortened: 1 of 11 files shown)


escowles:
@ruebot I've updated the README to be more explicit, and to also mention that the RDF format can be configured too.

bseeger:
Hi @escowles,

Am I doing this correctly: 

`java -jar ../../fcrepo4-labs/fcrepo-import-export/fcrepo-export/target/fcrepo-export-0.0.1-SNAPSHOT.jar --mode export --resource http://localhost:8080/fcrepo_4.6_rc/rest/ -binDir bin_dir --descDir desc_dir --rdfLang application/rdf+xml --rdfExt .xml`


The above results in a file named '.xml' in the directory for the root dir.  I also tried with the default turtle and got a '.ttl' file in that dir.  So, for me, starting at root results in an oddly named file.  The repo is relatively new and doesn't really have much in it yet.

escowles:
@bseeger I think the issue is the trailing / at the end of the URI — if you use `http://localhost:8080/fcrepo_4.6_rc/rest` you should get `desc_dir/rest.xml`

Assuming that works, then we probably need a ticket for stripping the trailing slash, so we handle that more gracefully.

bseeger:
@escowles - you're right, it was the trailing slash. I can file a ticket for this.

bseeger:
@escowles - should I be looking at a basic binary export at this point or is that beyond the scope of this PR?

awoods:
There is a lot of duplicate code here that should probably be pulled into a new fcrepo-import-export-common, but let's move this forward as-is for now, no?

barmintor:
@ruebot yes he did

PR #7, opened 2016-08-31
Binaries iff configured

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2141

### DIFF
--- a/fcrepo-export/src/main/java/org/fcrepo/export/Exporter.java
+++ b/fcrepo-export/src/main/java/org/fcrepo/export/Exporter.java
@@ -88,35 +88,44 @@ public class Exporter {
     }
     private void exportBinary(final URI uri)
             throws FcrepoOperationFailedException, IOException {
+        final File file = fileForBinary(uri);
+        if (file == null) {
+            logger.info("Skipping {}", uri);
+            return;
+        }
+
         try (FcrepoResponse response = client.get(uri).perform()) {
             logger.info("Exporting binary: {}", uri);
-            writeResponse(response, fileForBinary(uri));
+            writeResponse(response, file);
         }
     }
     private void exportContainer(final URI uri)
             throws FcrepoOperationFailedException, IOException {
+        final File file = fileForContainer(uri);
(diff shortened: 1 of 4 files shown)


fcrepo-export/src/main/java/org/fcrepo/export/Exporter.java
```diff
@@ -0,0 +1,140 @@
...
+        Writer w = null;
+        try {
+            if (!file.getParentFile().exists()) {
+                file.getParentFile().mkdirs();
+            }
+            w = new FileWriter(file);
+            copy(response.getBody(), w, "UTF-8");
+            logger.info("Exported {} to {}", response.getUrl(), file.getAbsolutePath());
+        } finally {
+            if (w != null) {
```
ajs6f: `Writer` < `AutoCloseable`, so can't we drop this `finally` by just using `try-with-resource`?

fcrepo-export/src/main/java/org/fcrepo/export/Exporter.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+    void writeResponse(final FcrepoResponse response, final File file)
+            throws IOException {
+        Writer w = null;
+        try {
+            if (!file.getParentFile().exists()) {
+                file.getParentFile().mkdirs();
+            }
+            w = new FileWriter(file);
+            copy(response.getBody(), w, "UTF-8");
```
ajs6f: This may be a later refinement, but consider using a reported size for the bitstream to select between `IOUtils::copy` and `::copyLarge`.

escowles:
@ruebot Do you have binaries in your repo?  I only see the jsonld files.

awoods:
This PR seems to export binaries iff `-b` is provided. I think what was confusing was the fact that the export is not recursive. When exporting `/parent` where `/parent/child` exists, only the `/parent` is exported.
Let's ship this PR and follow-on with:
https://jira.duraspace.org/browse/FCREPO-2142

PR #8, opened 2016-08-31
FCREPO-2151

barmintor:
The copy code ruined my day, so I branched the export-skeleton branch and started separating out some modules.


barmintor:
@ruebot rebased
https://jira.duraspace.org/projects/FCREPO/issues/FCREPO-2151

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ImportExportDriver.java
```diff
@@ -44,11 +48,11 @@ private void run(final String[] args) {
         final Config config = parser.parse(args);
 
         if (config.isImport()) {
-            final Importer importer = new Importer(config);
+            final Importer importer = new Importer((org.fcrepo.importer.Config)config);
```
awoods: I would have thought both the `Importer` and the `Exporter` could have accepted the `Config` interface as arguments instead of the package-specific classes... otherwise, what does the interface achieve?

PR #10, opened 2016-09-01
remove Config interface in favor of TransferProcess (FCREPO-2157)

barmintor:


awoods:
This looks good. I am inclined to merge it with the :+1: from at least one other person.

bseeger:
I think it looks good. I'd be disinclined to have duplicate code in both Config.java files (in fcrepo-export and fcrepo-import), but that was there before this, though it was tied together with the interface.  They are now just totally separate things.   Are there plans to try to move the config code so it can be more common?  Unless someone can imagine a scenario where it makes sense to keep them separate.

awoods:
It looks like the last PR broke the executable jar functionality. If we get this current PR into the codebase, I will fix the jar situation immediately.

awoods:
Did someone say integration test??

awoods:
@bseeger, I think @barmintor is suggesting that the two Configs are completely internal to Importer and Exporter, respectively; and therefore, will be completely independent. [TransferProcess](https://github.com/fcrepo4-labs/fcrepo-import-export/pull/10/files#diff-7ef0fe8a1b5d630730735c95158662ceR20) is the overlap.

PR #11, opened 2016-09-01
Fix broken executable jar

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2160


awoods:
Hit it! Then we can see if @escowles' PR now works.

PR #12, opened 2016-09-01
Removes trailing slash on resource, if present

bseeger:
-Added code to remove a trailing slash if one is present on the
URI of the resource that the user requested to export.

Resolves: https://jira.durspace.org/browse/FCREPO-2149

### DIFF
--- a/fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java
+++ b/fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java
@@ -108,14 +108,23 @@ public class Config {
      * Sets the URI of the resource to import/export
      */
     public void setResource(final String resource) throws URISyntaxException {
-        this.resource = new URI(resource);
+        if (resource != null && (resource.lastIndexOf('/') == resource.length() - 1)) {
+          this.resource = new URI(resource.substring(0, resource.length() - 1));
+        } else {
+          this.resource = new URI(resource);
+        }
     }
 
     /**
      * Sets the URI of the resource to import/export
      */
-    public void setResource(final URI resource) {
-        this.resource = resource;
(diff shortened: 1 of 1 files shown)


fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java
```diff
@@ -108,14 +108,23 @@ public File getDescriptionDirectory() {
      * Sets the URI of the resource to import/export
      */
     public void setResource(final String resource) throws URISyntaxException {
-        this.resource = new URI(resource);
+        if (resource != null && (resource.lastIndexOf('/') == resource.length() - 1)) {
```
escowles: Is there a reason not to use `resource.endsWith("/")` instead?  That seems more readable to me.

fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java
```diff
@@ -108,14 +108,23 @@ public File getDescriptionDirectory() {
...
+        }
     }
 
     /**
      * Sets the URI of the resource to import/export
      */
-    public void setResource(final URI resource) {
-        this.resource = resource;
+    public void setResource(final URI resource) throws URISyntaxException {
+        if (resource.toString().lastIndexOf('/') == resource.toString().length() - 1) {
```
escowles: Same question as above.

fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java
awoods: And maybe have this first `setResource` call the second `setResource(URI)`, and consolidate this substring logic in just one method.

awoods:
The code looks good. If the testing is good on your end, @escowles, please feel free to merge.

PR #14, opened 2016-09-01
Export binary descriptions

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2154

### DIFF
--- a/fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
+++ b/fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
@@ -83,7 +83,7 @@ public class Exporter implements TransferProcess {
             if(linkHeaders.contains(binaryURI)) {
                 exportBinary(uri);
             } else if (linkHeaders.contains(containerURI)) {
-                exportContainer(uri);
+                exportDescription(uri);
             } else {
                 logger.error("Resource is neither an LDP Container nor an LDP NonRDFSource: {}", uri);
             }
@@ -106,9 +106,9 @@ public class Exporter implements TransferProcess {
             writeResponse(response, file);
         }
     }
-    private void exportContainer(final URI uri)
+    private void exportDescription(final URI uri, final URI... describes)
             throws FcrepoOperationFailedException, IOException {
-        final File file = fileForContainer(uri);
+        final File file = fileForContainer(describes.length > 0 ? describes[0] : uri);
         if (file == null) {
             logger.info("Skipping {}", uri);
             return;
@@ -134,13 +134,18 @@ public class Exporter implements TransferProcess {
         }
     }
     void writeResponse(final FcrepoResponse response, final File file)
-            throws IOException {
+            throws IOException, FcrepoOperationFailedException {
         if (!file.getParentFile().exists()) {
             file.getParentFile().mkdirs();
         }
         try (Writer w = new FileWriter(file)) {
             copy(response.getBody(), w, "UTF-8");
             logger.info("Exported {} to {}", response.getUrl(), file.getAbsolutePath());
+        }
+
+        List<URI> describedby = response.getLinkHeaders("describedby");
+        if (describedby != null && describedby.size() > 0) {
+            exportDescription(describedby.get(0), response.getUrl());
         }
     }
     private File fileForBinary(final URI uri) {
--- a/fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
+++ b/fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
@@ -143,8 +143,8 @@ public class Exporter implements TransferProcess {
             logger.info("Exported {} to {}", response.getUrl(), file.getAbsolutePath());
         }
 
-        List<URI> describedby = response.getLinkHeaders("describedby");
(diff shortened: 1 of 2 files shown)


fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -106,9 +106,9 @@ private void exportBinary(final URI uri)
             writeResponse(response, file);
         }
     }
-    private void exportContainer(final URI uri)
+    private void exportDescription(final URI uri, final URI... describes)
             throws FcrepoOperationFailedException, IOException {
-        final File file = fileForContainer(uri);
+        final File file = fileForContainer(describes.length > 0 ? describes[0] : uri);
```
awoods: If there is more than one description, are we OK with ignoring all but the first?

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -134,14 +134,19 @@ private void exportMembers(final File file) {
...
             file.getParentFile().mkdirs();
         }
         try (Writer w = new FileWriter(file)) {
             copy(response.getBody(), w, "UTF-8");
             logger.info("Exported {} to {}", response.getUrl(), file.getAbsolutePath());
         }
+
+        List<URI> describedby = response.getLinkHeaders("describedby");
+        if (describedby != null && describedby.size() > 0) {
+            exportDescription(describedby.get(0), response.getUrl());
```
awoods: Similar comment as above for here.

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
escowles: I wouldn't be hard to handle multiple descriptions, so I'll go ahead and add support for that.

escowles: Though I don't think we can support multiple descriptions if we save them like the RDF serializer does — only one can have the same name as the binary.  So I think this would mean naming them based on the repository URI instead (e.g., `.../rest/file1/fcr:metadata.ttl` instead of `.../rest/file1.ttl`.

awoods: fhmm, if we drop descriptions, that seems problematic. As @mikedurbin [mentioned](https://wiki.duraspace.org/display/FF/Export+Format?focusedCommentId=77449413#comment-77449413), the `fcr:metatadata` syntax may not work on all platforms. 
Would an indexed form make sense:
`.../rest/file1.ttl`, `.../rest/file1-1.ttl`, `.../rest/file1-2.ttl`, etc? Or some other naming convention?

escowles: I definitely think we should serialize all of the descriptions.  Assuming we're talking only about current Fedora, then we're already doing that.  But since a future Fedora or another LDP implementation could have multiple descriptions, then handling multiples seems like a good idea to me.

AFAIK, the only platform that has issues with colons is MacOSX.  And filenames with colons work fine for me using the default filesystem type (HFS+ journaled).  We can always replace the colons with underscores or something to be sure it won't cause problems.

I think having the filenames based on the description URI also has the benefit of replacing the file when it's updated.  If we use numbered suffixes, then we wouldn't necessarily know which description went with which file and they might not get updated correctly because of that.

bseeger:
Wondering what, if anything, I'm doing wrong here?

`
BuckyKatt:/Users/bseeger/Projects/fedora/fcrepo4-labs/fcrepo-import-export/fcrepo-import-export-driver/target>java -jar fcrepo-import-export-driver-0.0.1-SNAPSHOT.jar -m export  --binDir ../../../../backup/export/bin_dir --descDir ../../../../backup/export/bin_dir --resource http://localhost:8080/fcrepo_4.6_rc/rest/
Exporting!
java.io.FileNotFoundException: ../../../../backup/export/bin_dir/fcrepo_4.6_rc/rest/goats/fcr_metadata.ttl (Not a directory)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:162)
	at java.io.FileWriter.<init>(FileWriter.java:90)
	at org.fcrepo.exporter.Exporter.writeResponse(Exporter.java:143)
	at org.fcrepo.exporter.Exporter.exportDescription(Exporter.java:119)
	at org.fcrepo.exporter.Exporter.writeResponse(Exporter.java:150)
	at org.fcrepo.exporter.Exporter.exportBinary(Exporter.java:107)
	at org.fcrepo.exporter.Exporter.export(Exporter.java:85)
	at org.fcrepo.exporter.Exporter.exportMembers(Exporter.java:130)
	at org.fcrepo.exporter.Exporter.exportDescription(Exporter.java:124)
	at org.fcrepo.exporter.Exporter.export(Exporter.java:87)
	at org.fcrepo.exporter.Exporter.run(Exporter.java:79)
	at org.fcrepo.importexport.ImportExportDriver.run(ImportExportDriver.java:52)
	at org.fcrepo.importexport.ImportExportDriver.main(ImportExportDriver.java:41)
`

PR #15, opened 2016-09-02
Using shared log config file

escowles:
Partially addresses https://jira.duraspace.org/browse/FCREPO-2153


escowles:
@ruebot it's a single commit, so you can go ahead and create a merge commit — thanks!

PR #16, opened 2016-09-02
Add integration test framework

awoods:
- Include test for testing executable jar

Resolves: https://jira.duraspace.org/browse/FCREPO-2155

### DIFF
--- a/fcrepo-import-export-driver/src/test/java/org/fcrepo/importexport/integration/ExecutableJarIT.java
+++ b/fcrepo-import-export-driver/src/test/java/org/fcrepo/importexport/integration/ExecutableJarIT.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


awoods:
@ruebot: Please, a ticket on the warnings would be great.

fcrepo-import-export-driver/src/test/java/org/fcrepo/importexport/integration/ExecutableJarIT.java
```diff
@@ -0,0 +1,113 @@
...
+                "-m", "export",
+                "-d", TARGET_DIR,
+                "-r", url.toString());
+
+        builder.directory(new File(TARGET_DIR));
+        final Process process = builder.start();
+        logger.debug("Output of jar run: {}", IOUtils.toString(process.getInputStream()));
+
+        // Verify
+        assertTrue("Process did not exit before timeout!", process.waitFor(1000, TimeUnit.SECONDS));
```
mikedurbin: As someone who is going to write more integration tests, the logic in lInes 91 through 103 should probably be a helper method.  You can already reuse that method in the other test.  This code will likely be used in almost every new test.  Then if we find we want to include stderr output in our logs (or something else), we'll only have to update one place.

Whether to include this assertion (or the next) is up to you... I can see how it might be messy to have assertions in a helper method, and I can envision tests in the future where we WANT an non-zero exitValue from the process.

mikedurbin:
I'm happy with this PR and it can be merged with or without addressing my above comment.  If not addressed, I'd expect the next added integration test to refactor code for optimal reuse.

PR #17, opened 2016-09-06
Updated exporter to allow for rdf and non-rdf in same output tree.

mikedurbin:
Sorry about all the added "final" statements and the resorting of imports: my IDE did it, and while it confuses the PR, it's probably not inappropriate.  If you thing it is, let me know and I'll fix it.


fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -147,7 +149,10 @@ void writeResponse(final FcrepoResponse response, final File file)
 
         final List<URI> describedby = response.getLinkHeaders("describedby");
         for (final Iterator<URI> it = describedby.iterator(); describedby != null && it.hasNext(); ) {
-            exportDescription(it.next());
+            exportDescription(response.getUrl(), it.next());
+            if (it.hasNext()) {
+                throw new RuntimeException("fcrepo does not produce multiple describedby headers.");
```
escowles: I had something very similar to this when I was first handling binary descriptions, but @awoods pointed out that Fedora (or other LDP impls) might have multiple binary descriptions.  So that's why I used the description URI instead of the binary URI to build the filename.  My thinking was that there could be multiple descriptions for a binary, but each description would need to have its own URI, so therefore could be used to build a unique filename.

ajs6f: Super 👍  to this. I cannot imagine we would put a upper cardinality restriction on descriptions in the API spec, either.

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -84,13 +85,13 @@ private void export(final URI uri) {
             if (linkHeaders.contains(binaryURI)) {
                 exportBinary(uri);
             } else if (linkHeaders.contains(containerURI)) {
-                exportDescription(uri);
+                exportDescription(uri, uri);
```
ajs6f: I think I get what's happening here, but this is so odd looking that it might be worth adding
```
/**
 * Some kind of explanation as to why this makes sense.
 */
private void exportDescription(final URI uri) { exportDescription(uri, uri); } 
```

escowles: Or make the `describesUri` parameter a vararg so it can be supplied or not depending on whether it was required.

mikedurbin:
Sorry, here https://jira.duraspace.org/browse/FCREPO-2165

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
mikedurbin: No, what happened there is that I sent the wrong commit:(  I flipped that one parameter to test my integration test.  I'll fix it...

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
mikedurbin: Ok, but I *realy* *really* *really* don't want our export format to be tied to the contentious hack that is "fcr:metadata".  Many of us had the understanding that fcr:metadata *is* an RDF representation of the binary.  Could I perhaps serialize fcr:metadata in the way I am, but serialize any other (not-currently-possible) descrbedBy resources as if they were normal resources?

ajs6f: Oh, "fcr:metadata" is purely an implementation convenience. You have to generate a URI for the first, required, description, and we do it by adding that string of characters. We don't want to privilege it in any way. We could do it (e.g.) by using a UUID, and sometimes I think we would do better to do something that opaque and thereby to force people to use the header as proper RESTafarians do. @mikedurbin is pointing to a deeper question (is a description only _related_ to its descriptand, or is it actually a _representation_ of it, or is it both?) that we probably want to try to drive around for this PR.

escowles: +1 to avoiding the semantic ditch of discussing the deeper issue of what the binary's description really is — I think that's great work in the API spec context, but we want working code this week, so we need to stay focused.

I think it's worth considering if we should inspect the RDF and use the subject of the triples to serialize the RDF to disk.  I don't know what you would do if there were two or more descriptions of the same URI, though.  The original suggestion was to just add a suffix to make them unique, so you'd have `/rest/file1.ttl`, `/rest/file1-1.ttl`, `/rest/file1-2.ttl`, etc.  That avoid overwriting the other descriptions, but then I don't know how you'd match descriptions to files if some of them were updated but not others.  Or if some disappeared.

I decided to use the request URL to generate the file path, reasoning that if we retrieve new RDF from the same URL, then we'd want to replace that file on disk.  This seems like a good 1-to-1 mapping that makes it easy to translate back and forth.  We will have files named after fcr:metadata because that's the URL in the describedby Link header, but it's not baked-in in any way.

ajs6f: Part of what's confusing is that we are trying to reproduce representations, but we are trying to do that by storing resources. So one way to cut this knot would be to _merge_ all of the descriptions to disk and then accept that while we can give you back all of your information, we cannot give you back the exact behavior of the API when you exported. That's not what we want, though, I don't think.

So the principle I kind of want to take forward is to make the serialization look as much as possible like the API, which is not a profound thought, but in this case it gives us practical guidance: descriptions should be stored _as separate resources_. That would point to @escowles choice: use the request URI.

mikedurbin: Alright... I'll make it work with the fcr:metadata URI.  Gimme a few minutes...

mikedurbin: I would like to register my complaint that building with an eye towards hypothetical future requirements at the expense of an elegant implementation against our current implementation often costs more time than it saves.

ajs6f: Whose time? Spent when? :) The whole idea is that spending a little time now will save a lot later.

mikedurbin: That's always the idea, but if it adds complexity (and it does in this case) and adds more unit tests (that cover behavior that fedora does not exhibit) every time we refactor those tests, change to new test libraries, run the build routine, etc. costs us time.  All for the unlikely potential that if/when we change the fedora spec and implementation we *might* not have to change this code as much?  Give me a break.  This code doesn't even do the things we need it to do now... maintaining code to allow it to do something that we couldn't exercise until the spec is complete and the implementation rewritten (or a new one made) seems a pretty pooruse of resources.

ajs6f: There is _nothing_ hypothetical about the requirements in hand. We have resources in hand now that have attached to them a technical characterization, triples that describe the relationships to other resources, MODS in XML, triples that describe their state of preservation, etc.... the use of multiple descriptions per resource is _exactly_ what we need _today_. The fact that we have to work on the problem from several fronts at once is annoying. I get that. But it's not a reason to just not work on the problem. If you want to work on the "repo-side" part of multiple descriptions per resource so that this part doesn't seem disconnected, rock on.

awoods:
@mikedurbin, it also seems reasonable to simply disallow both -d and -b being the same. No?

mikedurbin:
@awoods, if you don't have a desire to support having all your stuff serialized to a single directory with descriptions beside binaries it would be reasonable.  Presumably the issue reporter wanted to support that, and I certainly did, hence picking up the ticket and working towards a solution that attempt to meet everyone's vision.

I've issued another commit that avoids the conflict between a directory name and the filename and restored the support for a future fedora in which we have multiple describes links.  I added lambdas where others think they're more clear.  Let me know if I should make more changes.  I can rebase and squash if that makes review easier since the full diff is smaller than this most recent diff.

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -162,15 +150,14 @@ void writeResponse(final FcrepoResponse response, final File file)
 
         final List<URI> describedby = response.getLinkHeaders("describedby");
         for (final Iterator<URI> it = describedby.iterator(); describedby != null && it.hasNext(); ) {
```
ajs6f: I may be reading this wrong, but if `describedby` is `null`, isn't `describedby.iterator()` going to get executed anyway, throwing an NPE? That's based on what I can get out of https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.14.1.1

awoods:
@mikedurbin, It is a good question, @bseeger, do you actually want to store binaries and descriptions in the same hierarchy? or were you just doing good testing?

escowles:
I would like to store binaries and descriptions in the same directory — and having a different extension for binaries seems like a reasonably good way to do that.

ajs6f:
I don't think it's a terrible way to do that, but I do want to ask: how far are we willing to go to allow for "semi-custom" filesystem layouts? How much of a "center of gravity" do we want to create in the form of a standard, expect-able layout?

ajs6f:
I don't think it's a terrible way to do that, but I do want to ask: how far are we willing to go to allow for "semi-custom" filesystem layouts? How much of a "center of gravity" do we want to create in the form of a standard, expect-able layout?

ajs6f:
I don't think it's a terrible way to do that, but I do want to ask: how far are we willing to go to allow for "semi-custom" filesystem layouts? How much of a "center of gravity" do we want to create in the form of a standard, expect-able layout?

ajs6f:
I don't think it's a terrible way to do that, but I do want to ask: how far are we willing to go to allow for "semi-custom" filesystem layouts? How much of a "center of gravity" do we want to create in the form of a standard, expect-able layout?

fcrepo-export/src/main/java/org/fcrepo/exporter/ArgParser.java
```diff
@@ -73,6 +74,14 @@ public ArgParser() {
         binDirOption.setArgName("binDir");
         cmdOptions.addOption(binDirOption);
 
+        // binary extension option
+        final Option binExtOption = new Option("bx", "binExt", true, "Binary filename extension (default: " +
+                DEFAULT_BINARY_EXT);
+        binExtOption.setRequired(false);
+        binExtOption.setArgs(1);
+        binExtOption.setArgName("rdfExt");
```
awoods: Look like a typo. This should probably be `binExt`.

fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java
```diff
@@ -84,6 +85,20 @@ public File getBinaryDirectory() {
     }
 
     /**
+     * Sets the Binary filename extension.
+     */
+    public void setBinaryExtension(final String extension) {
```
awoods: [WARNING] Javadoc Warnings
[WARNING] /home/awoods/programming/java/apps/fedora/futures/fcrepo-import-export/fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java:96: warning: no @param for extension
[WARNING] public void setBinaryExtension(final String extension) {
[WARNING] ^
[WARNING] /home/awoods/programming/java/apps/fedora/futures/fcrepo-import-export/fcrepo-export/src/main/java/org/fcrepo/exporter/Config.java:103: warning: no @return
[WARNING] public String getBinaryExtension() {
[WARNING] ^

bseeger:
@awoods - I was just testing out variations.  While I like this feature (of putting both in the same dir tree), I would be fine with not being able to do that.

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
awoods: @ajs6f, yes, this looks like a potential NPE situation. Maybe a simple `if describedby != null` check would do the trick?

mikedurbin:
@awoods, I think this is good now.

mikedurbin:
Squshed and rebased.

fcrepo-import-export-common/src/main/java/org/fcrepo/importexport/Constants.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.importexport;
+
+/**
+ * @author Mike Durbin
+ */
+public abstract class Constants {
```
awoods: Would you be opposed to merging this class with `FcrepoConstants`?

mikedurbin: I'm not... I did it that way first, then reconsidered because the name of the file seemed misleading because the BINARY_EXTENSION constant isn't an "fcrepo" thing it's an "import-export" thing.  That said, it's an "fcrepo"-import-export thing.

fcrepo-import-export-driver/src/test/java/org/fcrepo/importexport/integration/ExecutableJarIT.java
```diff
@@ -89,20 +92,67 @@ public void testJarExport() throws Exception {
...
+                "-d", TARGET_DIR,
+                "-b", TARGET_DIR,
+                "-r", url.toString());
 
         // Verify
         assertTrue("Process did not exit before timeout!", process.waitFor(1000, TimeUnit.SECONDS));
         assertEquals("Did not exit with success status!", 0, process.exitValue());
 
-        assertTrue(new File(TARGET_DIR, url.getPath() + ArgParser.DEFAULT_RDF_EXT).exists());
+        response.getLinkHeaders("describedby")
```
awoods: We need to also assert that a `describedby` Link header exists.
If I replace `describedby` with `junk` in this test, it still passes.

fcrepo-import-export-common/src/main/java/org/fcrepo/importexport/Constants.java
awoods: Or if you prefer, you can change `FcrepoConstants` to `Constants`.

PR #20, opened 2016-09-06
Remove non-test dependencies on fcrepo

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2177


awoods:
We should not worry about those WARNINGs in this ticket.
I believe they are both non-issues.
...the ERRORs for missing package-list on the other hand would be nice to resolve:
```
[ERROR] Error fetching link: /home/awoods/programming/java/apps/fedora/futures/fcrepo-import-export/fcrepo-import-export-common/target/testapidocs/package-list
```

PR #25, opened 2016-09-08
Add support for config file

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2164

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ $ java -jar fcrepo-import-export-driver/target/fcrepo-import-export-driver-<vers
 ```
 
 The easiest way to see an example of the configuration file is to run the utility with command-line arguments and inspect the configuration file created.
-That configuration file will looks something like the following:
+That configuration file will look something like the following:
 > -m
 > export
 > -d
--- a/README.md
+++ b/README.md
@@ -8,11 +8,19 @@ Work in progress
 Requirements:
 * Java 8
 
-Additional requirements for building and running from source:
+Additional requirements for building:
 * Maven 3
 
+Modes of execution
+------------------
+The standalone import/export utility can be run in either of two ways:
+1) By passing in individual command-line arguments to the executable jar file
+2) By passing in a single configuration file that contains the standard command-line arguments
 
-To run the standalone import/export utility:
+The first time you run the utility with command-line arguments, a configuration file containing your provided arguments will be written to a file, the location of which will be displayed at the command line.
+
+Running the import/export utility with command-line arguments
+-------------------------------------------------------------
 
 ```sh
 $ java -jar fcrepo-import-export-driver/target/fcrepo-import-export-driver-<version>.jar --mode [import|export] [options]
@@ -33,3 +41,19 @@ java -jar fcrepo-import-export-driver/target/fcrepo-import-export-driver-0.0.1-S
 ```
 
 To then load that data into an empty Fedora repository at the same URL, run the same command, but using `--mode import` instead of `--mode export`.
+
+Running the import/export utility with a configuration file
+-----------------------------------------------------------
+
+```sh
+$ java -jar fcrepo-import-export-driver/target/fcrepo-import-export-driver-<version>.jar -c /path/to/config/file
+```
+
+The easiest way to see an example of the configuration file is to run the utility with command-line arguments and inspect the configuration file created.
+That configuration file will looks something like the following:
+> -m
+> export
+> -d
+> /tmp/rdf
+> -r
+> http://localhost:8080/rest/1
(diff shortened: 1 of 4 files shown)


escowles:
Can you update the README to include an example config file snippet and how to use the config file option?

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -37,78 +49,145 @@
  */
 public class ArgParser {
 
-    final private Options cmdOptions;
+    private static final Logger logger = getLogger(ArgParser.class);
 
     public static final String DEFAULT_RDF_EXT = ".ttl";
     public static final String DEFAULT_RDF_LANG = "text/turtle";
+    public static final String CONFIG_FILE_NAME = "config.txt";
```
escowles: Should we have a more descriptive config filename like `importexport.config`?  Do we expect this file to live in a directory along with the import-export JAR file, where it will be obvious what it is config for?

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -37,78 +49,145 @@
...
+    protected Config parseConfiguration(final String[] args) {
+        // Inspect Config File option
+        Config config = parseConfigFileOptions(args);
+        if (config == null) {
+            // Inspect standard command-line options
+            config = parseConfigurationArgs(args);
+        }
+
+        // Write configuration to disk
+        saveConfig(args);
```
escowles: Should this save the options as parsed in `config` instead of the original command-line params?  Wouldn't this lead to saving the pointer to the user-specified config file?

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -123,13 +202,42 @@ Config parseConfiguration(final String[] args) {
...
         return config;
     }
 
     /**
+     * This method writes the configuration file to disk
+     *
+     * @param args to be persisted
+     */
+    private void saveConfig(final String[] args) {
+        final File configFile = new File(System.getProperty("java.io.tmpdir"), CONFIG_FILE_NAME);
```
escowles: Is `java.io.tmpdir` a good place to store the config file?  Won't that be removed when the computer is rebooted?  Maybe it's a good fallback location for the config file (since we can be pretty sure it will exist and we can write a file there), but maybe some other user-configurable location would be better?

ajs6f: I don't understand why anything is getting put in the tmp directory here, but if you're going to do that, use [this](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempFile-java.lang.String-java.lang.String-java.nio.file.attribute.FileAttribute...-).

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -37,78 +49,145 @@
...
         final Option rdfLangOption = new Option("l", "rdfLang", true, "RDF language (default: " + DEFAULT_RDF_LANG);
         rdfLangOption.setRequired(false);
         rdfLangOption.setArgs(1);
         rdfLangOption.setArgName("rdfLang");
-        cmdOptions.addOption(rdfLangOption);
+        configOptions.addOption(rdfLangOption);
+
+        // Config file option
+        final Option configFileOption = new Option("c", "config", true, "Path to config file");
+        configFileOption.setRequired(true);
```
awoods: > Is the config option always required? 

No, you can pass in args or a config file.

> Why not allow specifying the required info on the command line instead?

Not sure I understand, what required info?

escowles: Shouldn't this be `setRequired(false)` to make it optional?

awoods: No, because it is required for the `configFileOptions`, not the `configOptions`.

README.md
```diff
@@ -33,3 +41,19 @@ java -jar fcrepo-import-export-driver/target/fcrepo-import-export-driver-0.0.1-S
...
+
+Running the import/export utility with a configuration file
+-----------------------------------------------------------
+
+```sh
+$ java -jar fcrepo-import-export-driver/target/fcrepo-import-export-driver-<version>.jar -c /path/to/config/file
+```
+
+The easiest way to see an example of the configuration file is to run the utility with command-line arguments and inspect the configuration file created.
+That configuration file will looks something like the following:
```
escowles: Typo: `looks` instead of `look`

escowles:
@awoods Looks good — do you want to squash or shall I squash & merge?

awoods:
@escowles, please squash/merge with the GitHub button.

PR #26, opened 2016-09-08
Added authorization support for import/export.

mikedurbin:
https://jira.duraspace.org/browse/FCREPO-2163

### DIFF
--- a/fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
+++ b/fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
@@ -40,7 +40,10 @@ import org.apache.jena.rdf.model.NodeIterator;
 import org.fcrepo.client.FcrepoClient;
 import org.fcrepo.client.FcrepoOperationFailedException;
 import org.fcrepo.client.FcrepoResponse;
+import org.fcrepo.importexport.AuthenticationRequiredRuntimeException;
+import org.fcrepo.importexport.AuthorizationDeniedRuntimeException;
 import org.fcrepo.importexport.Config;
+import org.fcrepo.importexport.ResourceNotFoundRuntimeException;
 import org.fcrepo.importexport.TransferProcess;
 import org.slf4j.Logger;
 
@@ -72,6 +75,9 @@ public class Exporter implements TransferProcess {
     }
 
     private FcrepoClient client() {
+        if (config.getUsername() != null) {
+            clientBuilder.credentials(config.getUsername(), config.getPassword());
+        }
         return clientBuilder.build();
     }
 
@@ -84,6 +90,7 @@ public class Exporter implements TransferProcess {
     }
     private void export(final URI uri) {
         try (FcrepoResponse response = client().head(uri).perform()) {
+            checkValidResponse(response, uri);
             final List<URI> linkHeaders = response.getLinkHeaders("type");
             if (linkHeaders.contains(binaryURI)) {
                 exportBinary(uri);
@@ -107,6 +114,7 @@ public class Exporter implements TransferProcess {
         }
 
         try (FcrepoResponse response = client().get(uri).perform()) {
+            checkValidResponse(response, uri);
             logger.info("Exporting binary: {}", uri);
             writeResponse(response, file);
         }
@@ -120,6 +128,7 @@ public class Exporter implements TransferProcess {
         }
 
         try (FcrepoResponse response = client().get(uri).accept(config.getRdfLanguage()).perform()) {
+            checkValidResponse(response, uri);
             logger.info("Exporting description: {}", uri);
             writeResponse(response, file);
         } catch ( Exception ex ) {
@@ -162,5 +171,27 @@ public class Exporter implements TransferProcess {
     private File fileForContainer(final URI uri) {
         return new File(config.getDescriptionDirectory(), TransferProcess.encodePath(uri.getPath())
             + config.getRdfExtension());
+    }
+
+    /**
+     * Checks the response code and throws a RuntimeException with a helpful
+     * message (when possible) for non 2xx codes.
+     * @param response the response from a REST call to Fedora
+     * @param uri the URI against which the request was made
+     */
+    private void checkValidResponse(final FcrepoResponse response, final URI uri) {
+        switch (response.getStatusCode()) {
+            case 401:
+                throw new AuthenticationRequiredRuntimeException();
+            case 403:
+                throw new AuthorizationDeniedRuntimeException(config.getUsername(), uri);
+            case 404:
+                throw new ResourceNotFoundRuntimeException(uri);
+            default:
+                if (response.getStatusCode() < 200 || response.getStatusCode() >= 300) {
+                    throw new RuntimeException("Export operation failed: unexpected status "
+                            + response.getStatusCode() + " for " + uri);
+                }
+        }
     }
 }
--- a/fcrepo-export/src/test/java/org/fcrepo/exporter/ExporterTest.java
+++ b/fcrepo-export/src/test/java/org/fcrepo/exporter/ExporterTest.java
@@ -21,8 +21,8 @@ import static org.fcrepo.importexport.FcrepoConstants.BINARY_EXTENSION;
 import static org.fcrepo.importexport.FcrepoConstants.CONTAINER;
 import static org.fcrepo.importexport.FcrepoConstants.CONTAINS;
 import static org.fcrepo.importexport.FcrepoConstants.NON_RDF_SOURCE;
-import static org.mockito.Mockito.eq;
-import static org.mockito.Mockito.isA;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.isA;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -31,14 +31,15 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.List;
 import java.util.Arrays;
+import java.util.List;
 
 import org.fcrepo.client.FcrepoClient;
 import org.fcrepo.client.FcrepoOperationFailedException;
 import org.fcrepo.client.FcrepoResponse;
-import org.fcrepo.client.HeadBuilder;
 import org.fcrepo.client.GetBuilder;
+import org.fcrepo.client.HeadBuilder;
+import org.fcrepo.importexport.AuthenticationRequiredRuntimeException;
 import org.fcrepo.importexport.Config;
 import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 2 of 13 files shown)


fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -163,4 +169,21 @@ private File fileForContainer(final URI uri) {
         return new File(config.getDescriptionDirectory(), uri.getPath().replaceAll(":", "_")
             + config.getRdfExtension());
     }
+
+    private void checkValidResponse(final FcrepoResponse response, final URI uri) {
```
ajs6f: I know this is `private`, but it might be nice to add the briefest doc here, just `@throws RuntimeException when authentication or authorization conditions fail` or the like.

fcrepo-export/src/test/java/org/fcrepo/exporter/ExporterTest.java
```diff
@@ -146,6 +152,18 @@ public void testExportContainer() throws Exception {
         Assert.assertTrue(exporter.wroteFile(new File("target/rdf/rest/1.jsonld")));
     }
 
+    @Rule
+    public ExpectedException expectedEx = ExpectedException.none();
```
ajs6f: This seems a bit oblique. Is there a reason that the one test that does throw an exception can't use the `expected` attribute on `@Test`?

mikedurbin: Nope... this is apparently the way people use JUnit to test whether an exception has a given message.  It's completely non-intuitive to me.  Would it be acceptable to catch the exception in the test method and do the comparison there with assert statements?

ajs6f: Ah, you need to examine the exception, I see. Yeah, I would actually prefer that you explicitly tangle out and assert about the exception in the test, because it's only one test. If there were a couple of different tests, I could see the savings in using the rule.

ajs6f: Hang on, I just looked at this a bit more carefully. Actually, I think the right thing to do here is subclass `RuntimeException` with something specific to the condition (`AuthenticationRequiredException`). Then your test code is typesafe _and_ you can use the simpler `expected` form, right? I know that adds a little more code, but throwing naked `RuntimeException`s is always a bit suspect to me.

mikedurbin: I can do that and agree it's better... I felt a little weird just doing it just for the purpose of testing.  There's lots of runtime exceptions thrown by the Importer and Exporter and I think a comprehensive look at them is warranted (which ones should halt operation, which ones can be logged without halting, whether the the messages should be written to log files or to standard out)  I thought the following ticket might address that, but it seems to be more limited: https://jira.duraspace.org/browse/FCREPO-2148

ajs6f: I get that (wanting to keep things simple for testing) but remember that Java is a strongly-typed language after all (for some meaning of that ambiguous descriptor) so it's almost never wrong to subtype. :) As to whether to do that here or in some other PR/ticket, that's certainly up to you and the Im/Ex sprinters. I'm not at all trying to block the code as-is, just offering suggestions.

fcrepo-export/src/main/java/org/fcrepo/exporter/Exporter.java
```diff
@@ -163,4 +170,27 @@ private File fileForContainer(final URI uri) {
...
     }
+
+    /**
+     * Checks the response code and throws a RuntimeException with a helpful
+     * message (when possible) for non 2xx codes.
+     * @param response the response from a REST call to Fedora
+     * @param uri the URI against which the request was made
+     */
+    private void checkValidResponse(final FcrepoResponse response, final URI uri) {
+        if (response.getStatusCode() >= 200 && response.getStatusCode() < 300) {
```
ajs6f: I'm just going on and on with this stuff, but I would rather read a `switch` here:
```
switch (response.getStatusCode()) {
    case 401:
        throw new AuthenticationRequiredRuntimeException();
    case 403:
        throw new RuntimeException("Export operation failed: authorization denied for " + config.getUsername());
    case 404:
        throw new RuntimeException("Export operation failed: resource not found " + uri);
    default:
         if (response.getStatusCode() < 200 || response.getStatusCode() => 300) {
             throw new RuntimeException("Export operation failed: unexpected status " + response.getStatusCode() + " for " + uri);
         }
}
```

fcrepo-import/src/main/java/org/fcrepo/importer/Importer.java
```diff
@@ -83,14 +87,10 @@ private FcrepoClient client() {
      */
     public void run() {
         logger.info("Running importer...");
-        try {
```
escowles: Why are you removing the exception handling here and converting the Exceptions below to RuntimeExceptions?  I think it's important to make sure the Exceptions get logged and don't just get passed up to the command line.

mikedurbin: They're already caught and logged by the driver.

https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ImportExportDriver.java#L47-L50

escowles: OK, so they bubble up to the driver, which logs them.  Then I'm 👍  on this PR — ready to merge, @ajs6f ?

ajs6f: 👍

escowles:
@mikedurbin Do you want to squash, or shall I squash & merge?

mikedurbin:
The merge conflicts will be a little rough... I'll take care of it and force push shortly.

fcrepo-export/src/test/java/org/fcrepo/exporter/ExporterTest.java
```diff
@@ -110,10 +112,12 @@ public void setUp() throws Exception {
         mockResponse(resource3, describedbyLinks, "binary");
         mockResponse(resource4, new ArrayList<>(), "{\"@id\":\"" + resource4.toString() + "\"}");
         mockResponse(resource5, new ArrayList<>(), "{\"@id\":\"" + resource5.toString() + "\"}");
+        mockResponse(resource5, new ArrayList<>(), "{\"@id\":\"" + resource5.toString() + "\"}");
```
awoods: This line should be removed.

fcrepo-import/src/main/java/org/fcrepo/importer/Importer.java
```diff
@@ -117,18 +117,20 @@ private void importFile(final File f) throws Exception {
...
             } else {
                 logger.info("Imported {}: {}", f.getAbsolutePath(), uri);
             }
         } catch (FcrepoOperationFailedException ex) {
-            throw new Exception("Error importing " + f.getAbsolutePath() + ": " + ex.toString(), ex);
+            throw new RuntimeException("Error importing " + f.getAbsolutePath() + ": " + ex.toString(), ex);
         } catch (IOException ex) {
-            throw new Exception("Error reading or parsing " + f.getAbsolutePath() + ": " + ex.toString(), ex);
+            throw new RuntimeException("Error reading or parsing " + f.getAbsolutePath() + ": " + ex.toString(), ex);
         } catch (URISyntaxException ex) {
```
awoods: We should not be throwing `URISyntaxExceptions`.
See: https://github.com/fcrepo4-labs/fcrepo-import-export/commit/fce76d6ad3bc00f1b53257f0093c3ad8d831d8f4

ajs6f: I disagree. It is useful to catch this (no one is throwing it) because it enables better diagnostic info than what you would get by just allowing 111 to go wrong. And given that arbitrary RDF could be going into this, 111 _could_ fail.

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -123,7 +132,18 @@ Config parseConfiguration(final String[] args) {
...
+                if (user.indexOf(':') == -1) {
+                    printHelp("user option must be in the format username:password");
+                } else {
+                    config.setUsername(user.substring(0, user.indexOf(':')));
+                    config.setPassword(user.substring(user.indexOf(':') + 1));
+                }
+            }
+
         } catch (ParseException e) {
+            e.printStackTrace();
```
awoods: We don't want to be `e.printStackTrace()`ing. If you want this stack trace, turn on DEBUG: https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ImportExportDriver.java#L49

fcrepo-import-export-driver/src/test/resources/repository-test.json
```diff
@@ -12,13 +12,13 @@
             "minimumBinarySizeInBytes" : 1
         }
     },
-    "security" : {
+        "security" : {
```
awoods: If you have a new commit coming, may as well remove this inadvertent space.

fcrepo-import/src/main/java/org/fcrepo/importer/Importer.java
awoods: Actually, someone is throwing it:
- https://github.com/fcrepo4-labs/fcrepo-import-export/pull/26/files#diff-d48f0429a55f2afa5045356359360d80R111
- https://github.com/fcrepo4-labs/fcrepo-import-export/pull/26/files#diff-d48f0429a55f2afa5045356359360d80R115

If you use `URI.create()` you get rid of the checked exception, but still get the same runtime exception detail:
https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#create(java.lang.String)

ajs6f: No one is throwing anything at those lines. More to the point, putting `URI::create` in there is, as I wrote, a bad idea because we lose the useful logging.

awoods: OK.

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -224,10 +267,15 @@ private void saveConfig(final String[] args) {
 
         // Write config to file
         try (final BufferedWriter configWriter = new BufferedWriter(new FileWriter(configFile));) {
-            for (final String arg : args) {
-                configWriter.write(arg);
-                configWriter.newLine();
-                configWriter.flush();
+            for (Option option : cmd.getOptions()) {
```
awoods: A comment would be helpful, noting that `user` arguments are being skipped.

PR #27, opened 2016-09-08
Translating URIs on import

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2180

### DIFF
--- a/fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
+++ b/fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
@@ -80,6 +80,14 @@ public class ArgParser {
                 .desc("Resource (URI) to import/export")
                 .required(true).build());
 
+        // Source Resource option
+        final Option sourceOption = new Option("s", "source", true,
+                "Source (URI) data was exported from, when importing to a different Fedora URI");
+        sourceOption.setRequired(false);
+        sourceOption.setArgs(1);
+        sourceOption.setArgName("source");
+        configOptions.addOption(sourceOption);
+
         // Binary Directory option
         configOptions.addOption(Option.builder("b")
                 .longOpt("binDir")
--- a/fcrepo-import/src/main/java/org/fcrepo/importer/SubjectMappingStreamRDF.java
+++ b/fcrepo-import/src/main/java/org/fcrepo/importer/SubjectMappingStreamRDF.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.importer;
+
+import static org.apache.jena.graph.Factory.createDefaultGraph;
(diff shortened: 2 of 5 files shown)


fcrepo-import/src/main/java/org/fcrepo/importer/SubjectMappingStreamRDF.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public void triple(final Triple t) {
+        sink.send(Triple.create(rebase(t.getSubject()), t.getPredicate(), rebase(t.getObject())));
+    }
+
+    private Node rebase(final Node node) {
+        if (node.isURI() && node.getURI().contains("/rest/")) {
+            return createURI(node.getURI().replaceFirst(".*/rest/", baseURI.toString() + "/"));
```
escowles: This isn't great (but it works given Fedora's typical URIs).  The other option I considered was having the client specify a second URI, and map from the first URI to the second URI.

fcrepo-import/src/main/java/org/fcrepo/importer/SubjectMappingStreamRDF.java
```diff
@@ -0,0 +1,85 @@
...
+        this.sink = new SinkTriplesToGraph(true, graph);
+    }
+
+    @Override
+    public void triple(final Triple t) {
+        sink.send(Triple.create(rebase(t.getSubject()), t.getPredicate(), rebase(t.getObject())));
+    }
+
+    private Node rebase(final Node node) {
+        if (node.isURI() && node.getURI().contains("/rest/")) {
```
acoburn: What if someone configures their repo such that URLs don't contain "rest"? (I.e. We have done this)

fcrepo-import/src/main/java/org/fcrepo/importer/SubjectMappingStreamRDF.java
ajs6f: That would be a good bit more flexible. Another approach would be to say, "Do your search-and-replace with sed or whatever on the filesystem."

escowles: Given that one of the main use cases is migration (where the base URI might reasonably change), I think we should support this in the client instead of punting to outside tools.  I'll add a config option for the baseURI to replace instead of depending on `/rest/`

ajs6f: 👍

escowles:
I just rebased to fix the options variable name that Travis was complaining about.

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -80,6 +80,13 @@ public ArgParser() {
         resourceOption.setArgName("resource");
         configOptions.addOption(resourceOption);
 
+        // Source Resource option
+        final Option sourceOption = new Option("s", "source", true, "Source (URI) data was exported from");
+        sourceOption.setRequired(false);
+        sourceOption.setArgs(1);
+        sourceOption.setArgName("resource");
```
awoods: Looks like a typo: `resource` -> `source`

awoods:
This looks good. It seems that the `source` command-line argument only applies to importing. I wonder if this argument (and any such others) should include in their `Option` description that it only has effect on import?

fcrepo-import-export-driver/src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -80,6 +80,13 @@ public ArgParser() {
         resourceOption.setArgName("resource");
         configOptions.addOption(resourceOption);
 
+        // Source Resource option
+        final Option sourceOption = new Option("s", "source", true, "Source (URI) data was exported from");
```
awoods: This looks good. It seems that the source command-line argument only applies to importing. I wonder if this argument (and any such others) should include in their Option description that it only has effect on import?

escowles:
@awoods I've updated the source option name to be "source", and the description to be "Source (URI) data was exported from, when importing to a different Fedora URI".

awoods:
@escowles, this is ready to go... shall I squash or you?

escowles:
@awoods I'll squash.

escowles:
@awoods Squashed!

PR #29, opened 2016-09-09
Fix exporting of external content resources

escowles:
Depends on https://github.com/fcrepo4-exts/fcrepo-java-client/pull/21
Fixes https://jira.duraspace.org/browse/FCREPO-2187


awoods:
This looks good, pending resolution of: fcrepo4-exts/fcrepo-java-client#21

escowles:
@ruebot I think that error is fixed by #30

PR #32, opened 2016-09-09
Adding import integration test

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2203


mikedurbin:
@escowles is this test meant to pass now, or only after we fix those bugs?

escowles:
@mikedurbin I think it will require fixing the ordering bug, I think it may be also showing the same errors that @bseeger mentioned with importing using auth.

escowles:
See https://travis-ci.org/fcrepo4-labs/fcrepo-import-export/builds/158834445#L4387

awoods:
Is there a related ticket to track this, however?

escowles:
@awoods I created FCREPO-2203 to track this.

PR #33, opened 2016-09-11
Refactored into two modules.

mikedurbin:
https://jira.duraspace.org/browse/FCREPO-2185


escowles:
Can we further collapse this to have the `fcrepo-import-export-driver/src` moved to `src`, and the `itests` be moved into `src/test/java/org/fcrepo/integration`?

mikedurbin:
I think so.  Do we want to change the name of the jar file to fcrepo-import-export and the documentation as well?

escowles:
@mikedurbin I think that would be nice.  In fact, we could probably completely remove the "driver" term by moving org.fcrepo.importexport.driver.ImportExportDriver to just org.fcrepo.importexport.ImportExport

mikedurbin:
@escowles it doesn't seem that pattern is allowed in maven: 'packaging' with value 'jar' is invalid. Aggregator projects require 'pom' as packaging.

awoods:
@escowles, I do not believe I follow the suggestion you are making in: https://github.com/fcrepo4-labs/fcrepo-import-export/pull/33#issuecomment-246342227
Can you elaborate?

awoods:
I do not believe the integration tests in `itests` are actually being run... at least I am not seeing evidence of their being run in the maven output.
Maybe it has to due with the packaging of `pom`: https://github.com/fcrepo4-labs/fcrepo-import-export/pull/33/files#diff-d578cebf69460f0ee4b3da730f6a7a57R13

escowles:
@awoods I'm suggesting that we should have a single module, so the top-level project would directly contain `src/main/`, `src/test/`, etc.

awoods:
@escowles, I thought @mikedurbin said a single module would not work due to dependency conflicts:
https://jira.duraspace.org/browse/FCREPO-2185?focusedCommentId=51824&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-51824

escowles:
@awoods Yes, @mikedurbin said that — and I'll defer to you on what is possible with Maven.  But I think I'm not understanding the problem.  It can't really be true that Maven doesn't support a single module with integration tests that produces a JAR artifact.  So I assume the problem is that we can't produce a JAR artifact and have integration tests in a separate module, or something like that.

ajs6f:
I think the issue raised is that Maven doesn't permit JAR projects with _submodules_.

escowles:
@ajs6f That makes sense — what I'm proposing is to not have any submodules, so I hope that should be fine.

mikedurbin:
So to summarize... we all think it'd be ideal to have no submodules, and have all the code in one simple project.  I punted on that because of some conflicts between our included ldpath libraries and libraries included by fedora.  (We could load fedora in a different way for integration tests, perhaps using the maven cargo plugin.)

So the next best option, was to have one module for all our code, and another for the integration tests (which includes fedora as dependencies).  This I implemented.

The suggestion by @escowles (if I understood correctly) was to have the parent module contain all our code and a submodule for integration tests.  This is not supported by maven.

ajs6f:
Two modules, one "main", one itest, isn't bad, but Cargo is probably best in the end.

awoods:
@mikedurbin, I agree, Maven allows a single module, or sub-modules... not both: top-level module and submodules. 
Believing the conflicts are true, I am fine with the two-module approach: `main` and `itests`.

escowles:
@mikedurbin OK, that clears things up.  I didn't mean to suggest having a single submodule for the integration tests, and now I see how that would prevent us from using JAR packaging for the main project.

I agree with @ajs6f that the current setup with two modules is an improvement, and that using Cargo would be even better (but is more work).

awoods:
@mikedurbin, do you have time/interest in exploring the cargo option?

mikedurbin:
Yeah... I'll give it a crack, if it's not holding anyone up for it to take a bit longer.

ajs6f:
My experience with Cargo is good. [This](https://codehaus-cargo.github.io/cargo/Passing+system+properties.html) is crucial.

ajs6f:
[Buildnumber](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/pom.xml#L449) is useful for ports for Cargo.

awoods:
@mikedurbin, not holding me up... as long as you feel like you can have something for review in the next day or two.

mikedurbin:
@awoods I'm struggling with the auth war files we distribute.  When you get a chance, could you run a quick test and let me know what I'm doing wrong.  Basically I've found that in my jetty cargo setup fcrepo-webapp-plus-webac-4.6.0.war does not expose anything under fcrepo/rest (404s).  The exact same setup for fcrepo-webapp-4.6.0.war works fine.  Furthermore, the fcrepo-webapp-plus-webac-4.6.0.war works fine in tomcat.  

I've set up two profiles in this PR to demonstrate.

mvn clean install 
mvn clean install -P auth

awoods:
@mikedurbin, I have been doing some digging into the `-Pauth` situation...
If you look at the maven output, you will see mention of where jetty is writing its log:
```
[INFO] [talledLocalContainer] 2016-09-13 22:06:44.280:INFO::main: Redirecting stderr/stdout to...
```
If you look in that log file, you will see that there is a Spring namespace handler issue:
```
Configuration problem: Unable to locate Spring NamespaceHandler for XML schema namespace [http://www.springframework.org/schema/p]
```
I am also seeing the same error if I deploy that war file into a local instance of jetty.

However, if I deploy that same war into a local instance of tomcat, all is well. Unfortunately, setting the jetty-plugin to use a tomcat container does not work because the tomcat plugin container does not download `http` war files like the jetty plugin container does... the tomcat plugin container expects local file paths.

It seems there are two options here:
1. Find a way to configure the jetty container plugin to correctly load in the spring namespace handlers (found in META-INF/spring.handlers files in individual jar files), or
2. Go back to the two module approach: `main` and `itests`

mikedurbin:
Thank you very much for helping me with this.

@awoods, that message doesn't appear in the log when I run it.  Did you have to change the log settings to have it appear?  

I found an easier option, which is to use another maven plugin to download the war file, then use cargo with tomcat7.  Using tomcat7 increases the build time by 30 seconds (or about 50%) on my machine, but I'm sure there's a command to skip those phases of the maven build.

ajs6f:
Just to be clear, @mikedurbin , the slowdown is because Tomcat is just bigger and heavier and takes longer to launch?

mikedurbin:
@ajs6f Yeah... jetty was built to be lightweight and more easily integrated into build processes.  Tomcat's kind of a hulk in by comparison.  That said, I suspect more people will run in production on Tomcat, so it's probably best to test against it.  (this is especially true since our "-auth" wars don't work in jetty)

awoods:
This is great, @mikedurbin. I will have this reviewed after my mid-day calls.

ajs6f:
That makes sense. Do we have a ticket for the fact that our auth-enabled gear doesn't work in Jetty? It's not that I think it so important in and of itself, but that not being portable across servlet containers is often a sign of an underlying fault.

awoods:
I suspect the Jetty issue is one of configuration. But a ticket would be great to track getting to the bottom of that.

pom.xml
```diff
@@ -18,10 +19,6 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
     <fcrepo-java-client.version>0.2.1-SNAPSHOT</fcrepo-java-client.version>
-    <!--We do NOT want to rely on a SNAPSHOT version of fcrepo; however, the dependency conflict between the-->
```
awoods: We should keep a modified version of this comment for the property: `fcrepo-java-client.version`.

pom.xml
```diff
@@ -43,12 +40,55 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+    </dependency>
+
+  </dependencies>
 
   <dependencyManagement>
```
awoods: With no sub-modules, we can remove both the `dependencyManagement` and `pluginManagement` sections.

mikedurbin:
I've addressed the comments so far, and made a ticket for the jetty issue.
https://jira.duraspace.org/browse/FCREPO-2210

src/test/resources/spring-test/noop.xml
```diff
@@ -6,21 +6,10 @@
...
-  <!-- Identifier translation chain -->
-  <util:list id="translationChain" value-type="org.fcrepo.kernel.api.identifiers.InternalIdentifierConverter">
-    <bean class="org.fcrepo.kernel.modeshape.identifiers.HashConverter"/>
-    <bean class="org.fcrepo.kernel.modeshape.identifiers.NamespaceConverter"/>
-  </util:list>
-
-  <context:component-scan base-package="org.fcrepo"/>
+  <!--  
+     Since we can't override the master spring configuration in the fcrepo-webapp.war artifact,
+     but we can override the included configurations for evening, jms and transactions, we
```
awoods: typo: `evening` to `eventing`

pom.xml
```diff
@@ -339,7 +221,92 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
       </plugin>
+
+      <plugin>
+        <groupId>com.googlecode.maven-download-plugin</groupId>
+        <artifactId>maven-download-plugin</artifactId>
+        <version>0.2-SNAPSHOT</version>
```
awoods: For consistency, may as well pull this version number into a property.

pom.xml
```diff
@@ -339,7 +221,92 @@
...
+              <targetDirectory>${project.build.directory}</targetDirectory>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.cargo</groupId>
+        <artifactId>cargo-maven2-plugin</artifactId>
+        <version>1.4.12</version>
```
awoods: For consistency, may as well pull this version number into a property.

pom.xml
```diff
@@ -1,234 +1,96 @@
...
+      <artifactId>logback-classic</artifactId>
+      <version>${logback.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
```
awoods: Removing this dependency results in a successful `mvn clean install`. Are we sure it is needed?

pom.xml
```diff
@@ -1,234 +1,96 @@
...
+      <artifactId>jackson-core</artifactId>
+      <version>${jackson.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.jsonld-java</groupId>
+      <artifactId>jsonld-java</artifactId>
```
awoods: Removing this dependency results in a successful `mvn clean install`. Are we sure it is needed?

pom.xml
```diff
@@ -1,234 +1,96 @@
...
+      <version>${jsonldjava.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
```
awoods: Removing this dependency results in a successful `mvn clean install`. Are we sure it is needed?

pom.xml
mikedurbin: Gosh, I wish someone (me) had scrutinized the imports more carefully before starting, I might not have had so much trouble reorganizing.   That said, I think we're better off with a project that avoids dependence on the fcrepo packages.

mikedurbin:
I'm happy to scrap imports that aren't necessary for the build to pass.  If for some reason they're needed, they can be added back with a test that demonstrates the need.

mikedurbin:
@awoods, good catch... I'm now using the latest version.

PR #34, opened 2016-09-15
Fixed basic import.

mikedurbin:
https://jira.duraspace.org/browse/FCREPO-2215

### DIFF
--- a/src/main/java/org/fcrepo/importexport/importer/Importer.java
+++ b/src/main/java/org/fcrepo/importexport/importer/Importer.java
@@ -67,6 +67,12 @@ public class Importer implements TransferProcess {
     protected FcrepoClient.FcrepoClientBuilder clientBuilder;
 
     /**
+     * The file, within the export directory for the directory mirrors the container
+     * to be imported (and whose content will be imported).
+     */
+    protected File importContainerDirectory;
+
+    /**
      * Constructor that takes the Import/Export configuration
      *
      * @param config for import
@@ -75,6 +81,10 @@ public class Importer implements TransferProcess {
     public Importer(final Config config, final FcrepoClient.FcrepoClientBuilder clientBuilder) {
         this.config = config;
         this.clientBuilder = clientBuilder;
+
+        if (config.getSource() == null) {
+            throw new RuntimeException("Source is required for import!");
+        }
     }
 
     private FcrepoClient client() {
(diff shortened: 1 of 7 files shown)


mikedurbin:
I was wrong, this also fixes: https://jira.duraspace.org/browse/FCREPO-2195

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -67,6 +67,12 @@
     protected FcrepoClient.FcrepoClientBuilder clientBuilder;
 
     /**
+     * The file, within the export directory for the directory mirrors the container
```
mikedurbin: Yeah, I'll try again, though this variable will mostly be useful in the future when we can export from one repo/path and import to another.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -89,49 +99,84 @@ private FcrepoClient client() {
...
     }
 
     private void importDirectory(final File dir) {
-        for (final File f : dir.listFiles()) {
-            if (f.isDirectory()) {
-                importDirectory(f);
-            } else if (f.isFile()) {
-                importFile(f);
-            }
+        for (final File f : dir.listFiles((f) -> {
```
mikedurbin: That is exactly what it does.... files before directories.  It looked clearer before checkstyle made me add all those line breaks to the lambdas.  I'll add a comment about the significance which will perhaps make it clearer.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -190,20 +236,17 @@ private InputStream modelToStream(final Model model) {
         return new ByteArrayInputStream(buf.toByteArray());
     }
 
-    private URI uriForFile(final File f, final File baseDir) throws URISyntaxException {
+    private String uriPathForFile(final File f, final File baseDir) throws URISyntaxException {
         String relative = baseDir.toPath().relativize(f.toPath()).toString();
-        if (relative.startsWith("rest") && config.getResource().toString().endsWith("/rest")) {
```
mikedurbin: Yeah, there's a test that prevents us from importing resources that have the rdf:type fedora:RepositoryRoot that I think takes care of this... that said... I had a lot of trouble understanding what this did.

src/main/java/org/fcrepo/importexport/importer/Importer.java
ajs6f: I thought we reset Checkstyle not to require pointless and confusing line breaks in lambdas?

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -89,49 +99,84 @@ private FcrepoClient client() {
...
+                    return;
+                }
+                final ResIterator binaryResources = model.listResourcesWithProperty(RDF_TYPE, NON_RDF_SOURCE);
+                if (binaryResources.hasNext()) {
+                    destinationUri = new URI(binaryResources.nextResource().getURI());
+                    logger.info("Importing binary {}", sourceRelativePath);
+                    response = importBinary(destinationUri, sanitize(model));
+                } else {
+                    destinationUri = new URI(config.getResource().toString()
+                            + (config.getResource().toString().endsWith("/") ? "" : "/")
```
mikedurbin: You're probably right... I always get confused when there's a semantic difference between including a slash at the end of the path and not.  In some posix program (rsync, for instance, I think) it's significant, but perhaps it never is in URL paths.

mikedurbin: Oops... that said, it looks like there's explicit logic in the setResource to ensure there's NOT a trailing slash.  I'll just add one without the inline logic.

mikedurbin:
I believe I've responded to all the comments, but it's not as clear to me from the new github comment system.

PR #36, opened 2016-09-19
Using OutputStreams instead of Writer to avoid corrupting binaries

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2218


src/test/java/org/fcrepo/importexport/integration/ExecutableJarIT.java
```diff
@@ -179,7 +179,9 @@ public void testExportBinaryAndRDFToSamePath() throws Exception {
         final File exportedBinary
                 = new File(TARGET_DIR, TransferProcess.encodePath(url.getPath()) + BINARY_EXTENSION);
         assertTrue(exportedBinary.exists());
-        assertEquals("Content was corrupted on export!", content, FileUtils.readFileToString(exportedBinary, "UTF-8"));
+        assert(exportedBinary.exists());
+        final String contentFromFile = new String(FileUtils.readFileToByteArray(exportedBinary));
+        assertEquals("Content was corrupted on export!", content, contentFromFile);
```
mikedurbin: This updated test passes even without your fix.

bseeger:
I tested this PR on two jpeg files and it works great.

escowles:
I've added a second commit that updates the test to fail before the update — loading the data from file was essential because Java won't let you have invalid UTF-8 characters in a Java source file.

PR #37, opened 2016-09-19
Creates dynamic ajp port for tomcat tests

bseeger:
Adds a dynamic port for the tomcat ajp port.  This will prevent
a collision if a tomcat server is already running on the machine.

Resolves: https://jira.duraspace.org/browse/FCREPO-2211


awoods:
@mikedurbin, can you verify this update?

mikedurbin:
Works for me!

PR #44, opened 2016-11-18
Adds  -h flag to display available options

dbernstein:
Resolves: https://jira.duraspace.org/browse/FCREPO-2196


src/main/java/org/fcrepo/importexport/ImportExportDriver.java
```diff
@@ -55,7 +55,6 @@ public static void main(final String[] args) {
     private void run(final String[] args) {
         final ArgParser parser = new ArgParser();
         final TransferProcess processor = parser.parse(args);
-
```
awoods: Please revert this format update.

PR #45, opened 2016-11-18
Removes -x (file extension) command line parameter.  The extension is…

dbernstein:
… implied by the

-l (language) parameter.  Also the ArgParser now throws an error on invalid language
params.

Resolves: https://jira.duraspace.org/browse/FCREPO-2200


src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -190,10 +190,10 @@ public URI getSource() {
     /**
      * Sets the RDF filename extension
      *
-     * @param extension of the RDF filename
+     * @param rdfExtension of the exported file
```
awoods: This same configuration is also used for import... which makes me believe the original javadoc is slightly more accurate.
Suggestion, revert all changes to this file.

awoods: Did you see the comment, @dbernstein ?

dbernstein: oops - I missed this one.

PR #48, opened 2016-12-05
DO NOT MERGE: Support mapping repository paths, not just host/port.

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2328

### DIFF
--- a/src/test/java/org/fcrepo/importexport/integration/AbstractResourceIT.java
+++ b/src/test/java/org/fcrepo/importexport/integration/AbstractResourceIT.java
@@ -25,22 +25,19 @@ import java.net.URI;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
-import org.apache.jena.query.Dataset;
 import org.apache.jena.rdf.model.Model;
-import org.apache.jena.sparql.core.DatasetImpl;
 import org.fcrepo.client.FcrepoClient;
+import org.fcrepo.client.FcrepoHttpClientBuilder;
 import org.fcrepo.client.FcrepoOperationFailedException;
 import org.fcrepo.client.FcrepoResponse;
 import org.junit.Before;
 import org.slf4j.Logger;
 
-import static org.apache.http.HttpStatus.SC_OK;
-import static org.apache.jena.graph.Node.ANY;
-import static org.apache.jena.graph.NodeFactory.createLiteral;
-import static org.apache.jena.graph.NodeFactory.createURI;
-import static org.apache.jena.rdf.model.ModelFactory.createDefaultModel;
+import static org.apache.jena.rdf.model.ResourceFactory.createPlainLiteral;
 import static org.apache.jena.rdf.model.ResourceFactory.createProperty;
-import static org.junit.Assert.assertEquals;
+import static org.apache.jena.rdf.model.ResourceFactory.createResource;
+import static org.apache.jena.riot.RDFDataMgr.loadModel;
+import static org.apache.jena.riot.web.HttpOp.setDefaultHttpClient;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
@@ -70,6 +67,7 @@ abstract class AbstractResourceIT {
 
     AbstractResourceIT() {
         clientBuilder = FcrepoClient.client().credentials(USERNAME, PASSWORD).authScope("localhost");
+        setDefaultHttpClient(new FcrepoHttpClientBuilder(USERNAME, PASSWORD, "localhost").build());
 
         final PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
         connectionManager.setMaxTotal(Integer.MAX_VALUE);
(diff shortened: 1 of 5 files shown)


src/test/java/org/fcrepo/importexport/integration/AbstractResourceIT.java
```diff
@@ -98,15 +95,22 @@ protected InputStream insertTitle(final String title) {
...
+    protected boolean exists(final URI uri) throws FcrepoOperationFailedException {
+        final FcrepoResponse resp = clientBuilder.build().head(uri).disableRedirects().perform();
+        return resp.getStatusCode() == 200 || resp.getStatusCode() == 307;
+    }
 
-        assertTrue(url + " should have had the dc:title, \"" + title + "\"!",
-                d.asDatasetGraph().contains(ANY, createURI(url.toString()),
-                        createProperty(DC_TITLE).asNode(), createLiteral(title)));
+    protected Model getAsModel(final URI uri) throws FcrepoOperationFailedException {
+        final FcrepoResponse getResponse = clientBuilder.build().get(uri).accept("text/turtle").perform();
```
ajs6f: [Yep.](https://jena.apache.org/documentation/query/http-auth.html) You would use `HttpOp.setDefaultHttpClient(HttpClient httpClient)`. In fact, I was just working on that in Jena this past summer, funny enough.

escowles: Cool — much tidier to use RDFDataMgr than to fetch and parse ourselves.  I've pushed a commit to do that.

awoods:
This looks and works good, @escowles. I was about to squash and merge... but how would you like to handle that?

escowles:
@awoods I've gone ahead and squashed my commits.

awoods:
@escowles , In considering this further, I am unclear of the source of the requirement that import/export should support exporting a resource then importing it to a different location (i.e. as a different resource). This works fine for single resources or resources that are not inter-referenced... but for any realistic repository, this functionality would appear to be opening the door to import failures.
Can you please remind me of the need or requirement here?

escowles:
@awoods The primary use case I had in mind was if the repository path changes (e.g. `/rest` to `/fcrepo/rest`).  But other changes in the path seem reasonable to me.  For example, Hydra typically adds a top-level container for each Rails environment, so resources would be in `/rest/dev` or `/rest/prod`.  Handling these kinds of changes to the baseURL seems useful, and I was surprised that it didn't already work when I tried it.

It's also convenient for adding ITs for https://jira.duraspace.org/browse/FCREPO-2192 — we could start and stop the repository, but it seems much easier to export from one path and import into another path.

Though now that you bring it up, I agree that the translation might not handle resources that are not children of the exported resource.  For example, if you export `/rest/dev/a` and it references `/rest/dev/b`, then translating `/rest/dev/a` to `/rest/prod/a` would likely fail.  It may be possible to handle this case too — let me see if I can find a mapping that will handle that gracefully.

escowles:
@ruebot Yep, I'll rebase and see if I can address @awoods' concerns — should have more info by our call today.

awoods:
@escowles , Thanks for looking more into this. I have also successfully tested the Hydra use case, i.e. export from `/rest/dev` and import to `/rest/prod`, where:
```
</rest/dev/a> dc:hasPart </rest/dev/b>
```
...which is comforting.

The caveat with the capability offered by this PR is that resources that have been exported from a given repository sub-tree can not reference resources found outside of that same sub-tree. Otherwise, this seems good.

escowles:
@awoods If the final part of the source and destination URIs is the same, then I think we could update the mapping algorithm to remove the matching parts from the source and destination URIs to determine the correct base URIs to use for mapping.

E.g., if you export `http://localhost:8080/rest/dev/a/b/c/abc123` and then import it to `http://localhost:8080/rest/prod/a/b/c/abc123`, then we should be able to remove the `/a/b/c/abc123` from the ends to figure out that we should map `/rest/dev` to `/rest/prod`.

Does that seem like a reasonable approach?

awoods:
@escowles , I must not be understanding your example. Because, I believe that scenario currently works, no?

escowles:
@awoods It works as long as long as `http://localhost:8080/rest/dev/a/b/c/abc123` doesn't reference other resources like `http://localhost:8080/rest/dev/d/e/f/def456`.  I mean that it should be able to figure out that `/rest/dev` is the base path and translate `http://localhost:8080/rest/dev/d/e/f/def456` to `http://localhost:8080/rest/prod/d/e/f/def456`.

awoods:
hmm... I still must not be understanding, because I just tested the scenario you describe... and import/export works.
```
</rest/dev/a/b/c/abc123> dc:hasPart </rest/dev/d/e/f/def456>
```
- export `/rest/dev`
- import to `/rest/prod`
Now see:
```
</rest/prod/a/b/c/abc123> dc:hasPart </rest/prod/d/e/f/def456>
```

escowles:
@awoods I think the key part is not exporting/importing the baseDir, but a child directory:

```
</rest/dev/a/b/c/abc123> dc:hasPart </rest/dev/d/e/f/def456>
```

export `/rest/dev/a/b/c/abc123`
import to `/rest/prod/a/b/c/abc123`

awoods:
Got it. Thanks. 
That is also what I was trying to communicate here: https://github.com/fcrepo4-labs/fcrepo-import-export/pull/48#issuecomment-265475718

The question is, what would we want the `dc:hasPart` relationship to be after import in this scenario?
1. `</rest/prod/a/b/c/abc123> dc:hasPart </rest/prod/d/e/f/def456>` --noting that the object of this triple may likely not exist
2. `</rest/prod/a/b/c/abc123> dc:hasPart </rest/dev/d/e/f/def456>`
??

escowles:
@awoods I would think the former (`</rest/prod/a/b/c/abc123> dc:hasPart </rest/prod/d/e/f/def456>`) because if you exported the entire `/rest/dev` tree, that resource would be created at that URI.  Since it doesn't exist, we'd create a placeholder for it.

awoods:
@escowles , that sounds reasonable... but it also sounds like we are tabling this as a "Phase-3" requirement.

escowles:
Closing this PR since we've decided not to pursue this functionality right now

PR #49, opened 2016-12-06
First pass at an audit logger.

whikloj:
Related to [FCREPO-2169](https://jira.duraspace.org/browse/FCREPO-2169), [FCREPO-2170](https://jira.duraspace.org/browse/FCREPO-2170) and sort of  [FCREPO-2237](https://jira.duraspace.org/browse/FCREPO-2237)


whikloj:
@ruebot @awoods Please have a look at this when you've got a chance. This seems cleaner, it doesn't generate the `null` directory of logs during tests. It basically just uses some logback trickery to get the changing filename and default directory.

src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -17,20 +17,8 @@
  */
 package org.fcrepo.importexport;
 
-import org.apache.commons.cli.CommandLine;
```
escowles: It looks like  only whitespace/ordering changes are left in this file — can you revert them?

src/main/resources/logback.xml
```diff
@@ -6,6 +6,17 @@
...
     </encoder>
   </appender>
+
+  <timestamp key="bySecond" datePattern="yyyyMMdd'T'HHmmss"/>
+  <appender name="AUDIT_FILE" class="ch.qos.logback.core.FileAppender">
+    <!-- use the previously created timestamp to create a uniquely
+         named log file -->
+    <file>${fcrepo.log.importexport.logdir:-${PWD}}/importexport_audit_${bySecond}.log</file>
+    <encoder>
+      <pattern>%date %level %logger{10} %msg%n</pattern>
```
awoods: I would suggest changing the `pattern` to:
```
<pattern>%date %msg%n</pattern>
```

awoods:
Nice work on the first pass.
It would be good to get some feedback from the sprint team on these two items:
1. Should we add an option flag that allows the user to enable/disable the audit log, with the default being `off`?
2. Should we add an option for specifying the output directory of the audit log? (This would be particularly nice during builds, since my project directory is filling up with log files generated during the build process)

whikloj:
So you can set the directory to stick the file in with the `fcrepo.log.importexport.logdir` property. Ala `java -Dfcrepo.log.importexport.logdir=/my/log/directory -jar ....`

However if you want to turn it off, I'll probably have to go back to configuring it in the application.

Last issue is related to the counter, we export the root resource but we don't import it. So in a test export and import I am missing one item.
```
==> importexport_audit_20161207T184943.log <==
2016-12-07 18:53:06,361 Finished export... 24258 resources exported

==> importexport_audit_20161207T202904.log <==
2016-12-07 20:38:24,779 Finished import... 24257 resources imported
```

whikloj:
I'm not super hot about the NullLogger class, but otherwise I would have to go back to creating a Logger in code. 
The SiftingAppender stops file creation until you actually send a log message and the option between the real logger and the NullLogger allows you to shut it right off.

src/main/java/org/fcrepo/importexport/common/NullLogger.java
```diff
@@ -0,0 +1,396 @@
+/*
```
escowles: Is this different from NOPLogger?  http://www.slf4j.org/api/org/slf4j/helpers/NOPLogger.html

src/main/java/org/fcrepo/importexport/common/NullLogger.java
```diff
@@ -0,0 +1,396 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.Marker;
+
+/**
+ * Do nothing, absolutely nothing.
+ *
+ * @author whikloj
+ */
+public class NullLogger implements Logger {
```
awoods: yuck

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -77,6 +78,11 @@ public Exporter(final Config config, final FcrepoClient.FcrepoClientBuilder clie
         this.clientBuilder = clientBuilder;
         this.binaryURI = URI.create(NON_RDF_SOURCE.getURI());
         this.containerURI = URI.create(CONTAINER.getURI());
+        if (config.getAuditLog()) {
```
awoods: Could you instead have `config` return the appropriate logger?

src/main/java/org/fcrepo/importexport/common/NullLogger.java
awoods: It looks like NOPLogger would do the trick.

whikloj:
Thank you @escowles, I figured there had to be someone out there who needed to log nothing too.

whikloj:
Rebased
I have been testing with these 4 commands.
`java -jar target/fcrepo-import-export-0.0.1-SNAPSHOT.jar`
Get an error no audit log produced

`java -jar target/fcrepo-import-export-0.0.1-SNAPSHOT.jar -a`
Get an error no audit log produced

`java -jar target/fcrepo-import-export-0.0.1-SNAPSHOT.jar -m export -b temp_output -b -r http://localhost:8080/rest`
Get export no audit log produced

`java -jar target/fcrepo-import-export-0.0.1-SNAPSHOT.jar -m export -b temp_output -b -r http://localhost:8080/rest -a`
Get export with audit log produced.

Good to know if I am counting correctly, I get confused about what counts in some of these situations.

escowles:
@ruebot This looks good to me — go ahead and take it for a spin

PR #52, opened 2016-12-08
Supporting custom predicates for defining resource containment

escowles:
* Allows specifying one or more predicates to define resource containment.
* By default ldp:contains is used

Fixes https://jira.duraspace.org/browse/FCREPO-2225


awead:
@ruebot sure

awead:
@escowles confirmed! I Tested ldp:contains and pcdm:hasMember predicates with the test hydra dataset. Ex:

``` bash
java -jar target/fcrepo-import-export-0.0.1-SNAPSHOT.jar  \
  -d hydra-export -u fedoraAdmin:fedoraAdmin \
  -r http://localhost:8986/rest/dev/qv/33/rw/64/qv33rw64w -m export \
  --predicate http://pcdm.org/models#hasMember,http://www.w3.org/ns/ldp#contains
```

PR #53, opened 2016-12-09
Implements binary redirects export approach

dbernstein:
Resolves https://jira.duraspace.org/browse/FCREPO-2296

### DIFF
--- a/src/test/java/org/fcrepo/importexport/importer/ImporterTest.java
+++ b/src/test/java/org/fcrepo/importexport/importer/ImporterTest.java
@@ -48,23 +48,31 @@ public class ImporterTest {
     private FcrepoClient client;
     private FcrepoClient.FcrepoClientBuilder clientBuilder;
     private Config binaryArgs;
-    private Config noBinaryArgs;
+    private Config externalResourceArgs;
     private Config containerArgs;
     private Config pairtreeArgs;
     private URI binaryURI;
     private URI binaryDescriptionURI;
+    private URI externalResourceURI;
+    private URI externalResourceDescriptionURI;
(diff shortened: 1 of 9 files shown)


src/test/java/org/fcrepo/importexport/importer/ImporterTest.java
```diff
@@ -144,8 +165,19 @@ public void testImportBinary() throws Exception {
...
+        verify(client).put(externalResourceURI);
+        verify(externalResourceBuilder).body(eq(new File(externalFilesDir, "rest/ext1.external")),
+                                eq("message/external-body"));
+        verify(client).put(externalResourceDescriptionURI);
+    }
+
+    @Test
     public void testImportWithoutBinaries() throws Exception {
-        final Importer importer = new Importer(noBinaryArgs, clientBuilder);
+        binaryArgs.setIncludeBinaries(false);
```
escowles: This looks like a good way to avoid a bunch of redundant objects above — but are you sure the tests are run in a predictable order and this won't interfere with the other tests that expect it to include binaries?

escowles: After some brief discussion on IRC, I think that the order the tests run in is dependent on the JVM that's running them.  So please make another copy of the config instead of updating it in the test.

dbernstein: For clarity I'm happy to use a copy of the config.  But in terms of order,  it shouldn't matter considering the binaryArgs flag gets set to true in the setup() method which runs before each test.  But you're right, it could open the door for confusion down the road.

PR #54, opened 2016-12-09
Setting digest header on uploaded binaries to detect corruption

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2334


awead:
@escowles Ok, this looks good. I've verified that if I either corrupt the binary from an export, or change the checksum, I get:
```
ERROR 15:28:57.093 (ImportExportDriver) Error performing import/export: Index: 0, Size: 0
```
Some possible improvements:
* better error message
* import the other resources after encountering a failure

I have three binaries in my export. The import process retrieves them in tree order. I corrupted the second one in the list and the import process stopped and didn't get to the third. Maybe it should? Otherwise, this is all good to merge.

escowles:
@awead I've pushed a new commit that improves the logging of failures, and has the import continue after failures.

escowles:
Rebasing after #58 was merged.

dbernstein:
I tested it.  Import with good sha1 succeeds, while bad one fails as expected.

PR #55, opened 2016-12-09
Don't try to parse binaries as RDF

escowles:
* Skip parsing binaries and any other non-RDF files
* Don't swallow IO/RDF parsing errors

Fixes https://jira.duraspace.org/browse/FCREPO-2330


whikloj:
Makes sense, was this trying to parse those binaries failing and then continuing before?

escowles:
@whikloj Yes, it was catching Exception so any error trying to parse the files was being caught and logged, but not stopping the import.

PR #58, opened 2016-12-09
Use Yaml for config file and put defaults in Config

whikloj:
Resolves: https://jira.duraspace.org/browse/FCREPO-2340

Also moves the definition of defaults inside the Config object.


jwestgard:
I'm seeing an empty yaml file being created when I run the exporter. Has anyone else seen this?

whikloj:
Looking into it.

whikloj:
@ruebot @jwestgard My mistake. This should be better.

jwestgard:
Confirmed -- getting a good yaml file now.

jwestgard:
No, I'm not a member of this one.

PR #63, opened 2016-12-13
Initial Bag profile implementation

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2354


escowles:
@awead: There are tickets for the Bag profiles:

https://jira.duraspace.org/browse/FCREPO-2222
https://jira.duraspace.org/browse/FCREPO-2223
https://jira.duraspace.org/browse/FCREPO-2224

I didn't fully review the comments in those tickets, so I probably need to update the profiles I created in src/main/resources/profiles with info from those.

escowles:
There's another ticket for a Bag profile for Perseids: https://jira.duraspace.org/browse/FCREPO-2306 — that ticket has a separate spec for tag manifests (so I think we should be using that instead of just assuming they are the same as the payload manifests).

awead:
👍 merge it!

PR #65, opened 2016-12-14
Adds support for user-supplied metadata

dbernstein:
Resolves: https://jira.duraspace.org/browse/FCREPO-2352

### DIFF
--- a/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
+++ b/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
@@ -57,6 +57,7 @@ import org.fcrepo.client.FcrepoOperationFailedException;
 import org.fcrepo.client.FcrepoResponse;
 import org.fcrepo.importexport.common.AuthenticationRequiredRuntimeException;
 import org.fcrepo.importexport.common.AuthorizationDeniedRuntimeException;
+import org.fcrepo.importexport.common.BagConfig;
 import org.fcrepo.importexport.common.BagProfile;
 import org.fcrepo.importexport.common.Config;
 import org.fcrepo.importexport.common.ResourceNotFoundRuntimeException;
@@ -121,6 +122,11 @@ public class Exporter implements TransferProcess {
             try {
                 // parse profile
                 final URL url = this.getClass().getResource("/profiles/" + config.getBagProfile() + ".json");
+
+
+                final BagConfig bagConfig = loadBagConfig(config.getBagConfigPath());
+
+
                 final InputStream in = (url == null) ? new FileInputStream(config.getBagProfile()) : url.openStream();
                 final BagProfile bagProfile = new BagProfile(in);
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -129,6 +135,8 @@ public Exporter(final Config config, final FcrepoClient.FcrepoClientBuilder clie
                 bagdir.mkdirs();
                 this.bag = new Bag(new Version(0, 97));
                 this.bag.setRootDir(bagdir);
+                this.bag.getMetadata().putAll(bagConfig.getBagInfo());
+                //TODO where to add aptrust bag info from the bagConfig?
```
escowles: I don't know if there is support for custom tag files in bagit-java — I looked at the API docs and don't see it anywhere.  So we may have to write our own.  Luckily, the format is just

```
key: value
```

So it shouldn't be hard to just iterate over the APTrust values and output them.

dbernstein: @escowles :  so are you suggesting that we just write in our own aptrust-info.txt filling it with kv pairs from the bagConfig.getAPTrustInfo()?

src/test/resources/configs/bagit-config.yml
```diff
@@ -0,0 +1,14 @@
...
+  Contact-Name: My Name
+  Contact-Phone: +19999999999
+  Contact-Email: anyone@university.edu
+  External-Description: ???
+  External-Identifier: ???
+  Bag-Group-Identifier: ???
+  Internal-Sender-Identifier: ???
+  Internal-Sender-Description: ???
+aptrust-info.txt:
+  Access: something
```
dbernstein: will do.

src/test/java/org/fcrepo/importexport/common/BagConfigTest.java
```diff
@@ -0,0 +1,47 @@
...
+public class BagConfigTest {
+
+    @Test
+    public void testFromFile() throws Exception {
+        final File testFile = new File("src/test/resources/configs/bagit-config.yml");
+        final BagConfig config = new BagConfig(testFile);
+
+        final Map<String,String> bagInfo = config.getBagInfo();
+        assertNotNull(bagInfo);
+        assertNotNull(bagInfo.get(BagConfig.SOURCE_ORGANIZATION_KEY));
```
dbernstein: will do.

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
escowles: Yes, for reference, that's what BagWriter is doing: https://github.com/LibraryOfCongress/bagit-java/blob/master/src/main/java/gov/loc/repository/bagit/writer/BagWriter.java#L255

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -129,6 +135,8 @@ public Exporter(final Config config, final FcrepoClient.FcrepoClientBuilder clie
                 bagdir.mkdirs();
                 this.bag = new Bag(new Version(0, 97));
                 this.bag.setRootDir(bagdir);
+                this.bag.getMetadata().putAll(bagConfig.getBagInfo());
```
escowles: @dbernstein: I've created a new PR that adds a fix for this to your commits: https://github.com/fcrepo4-labs/fcrepo-import-export/pull/66

PR #66, opened 2016-12-15
Adds support for user-supplied metadata

escowles:
* Alternative to #65 
* Avoid overwriting user-supplied metadata with tech metadata

Fixes https://jira.duraspace.org/browse/FCREPO-2352

### DIFF
--- a/src/main/java/org/fcrepo/importexport/ArgParser.java
+++ b/src/main/java/org/fcrepo/importexport/ArgParser.java
@@ -57,6 +57,16 @@ import com.esotericsoftware.yamlbeans.YamlWriter;
  */
 public class ArgParser {
 
+    /**
+     *
+     */
+    private static final String BAG_CONFIG_OPTION_KEY = "bag-config";
+
+    /**
+     *
+     */
+    private static final String BAG_PROFILE_OPTION_KEY = "bag-profile";
+
     private static final Logger logger = getLogger(ArgParser.class);
 
     public static final String CONFIG_FILE_NAME = "importexport.yml";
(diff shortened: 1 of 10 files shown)


src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -189,13 +206,33 @@ protected Config parseConfiguration(final String[] args) {
...
 
     /**
+     * Validation routines for checking interdependent configuration parameters.
+     * @param config
+     */
+    private void validateConfig(final Config config) {
+        // verify bagit config
+        if (config.getBagProfile() != null &&
+                config.getBagConfigPath() == null) {
+            throw new RuntimeException("A bagit config path must be set when using a bagit profile.");
```
whikloj: Do we need to set a tag config file for every run? Could we see if the profile has required tags and then check to see if those requirements are met?

escowles: In theory, you could have a profile that didn't require any user-supplied metadata.  Even our default profile is requiring the organization name, but you could have a profile called "minimal" that would export valid bags with no extra metadata at all.  So having the option to skip this if there were no required fields seems reasonable.

whikloj: See I didn't realize that because up till now we haven't been enforcing those restrictions so I thought this was adding an additional layer of complexity. I still think if the profile doesn't require extra tags we shouldn't require a person to make a config file. But I also don't like the idea of another option to turn this off. So I'll change my vote and say leave it as is.

PR #67, opened 2016-12-15
Enforces BagIt profile requirements for user-supplied metadata.

dbernstein:
Resolves https://jira.duraspace.org/browse/FCREPO-2356


src/test/java/org/fcrepo/importexport/exporter/ExporterTest.java
```diff
@@ -145,6 +147,15 @@ public void setUp() throws Exception {
         when(headResponse.getStatusCode()).thenReturn(200);
     }
 
+    @After
+    public void tearDown() {
+        try {
+            FileUtils.deleteDirectory(new File(exportDirectory));
```
whikloj: Probably don't need this, files are in the target directory and will get cleaned with `mvn clean`.

dbernstein: Some unit tests were failing when this wasn't in place.  Downstream tests check for the absence of certain files.

dbernstein:
@ruebot :  re your comment above: yes - I'll fix that.

PR #69, opened 2016-12-16
update version to 0.1.0-SNAPSHOT

acoburn:
Typically, the first released version of an artifact would be 0.1.0 (as opposed to 0.0.1). This updates the current snapshot value to reflect that.


acoburn:
@ruebot feel free to wait on this then (or close it). I would only recommend that the first release not be 0.0.1 but rather 0.1.0.

PR #73, opened 2016-12-16
Only validate BagIt fields for active profile

whikloj:
Resolves [FCREPO-2365](https://jira.duraspace.org/browse/FCREPO-2365)

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/BagConfig.java
+++ b/src/main/java/org/fcrepo/importexport/common/BagConfig.java
@@ -148,9 +148,6 @@ public class BagConfig {
      * @return a map of filenames to key-value property maps
      */
     public Map<String, String> getFieldsForTagFile(final String tagFile) {
-        if (map.get(tagFile) != null) {
-            return map.get(tagFile);
-        }
-        return null;
+        return map.get(tagFile);
     }
 }
--- a/src/main/java/org/fcrepo/importexport/common/BagConfig.java
+++ b/src/main/java/org/fcrepo/importexport/common/BagConfig.java
@@ -23,10 +23,8 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
-
-import org.apache.commons.lang3.StringUtils;
+import java.util.Set;
 
 import com.esotericsoftware.yamlbeans.YamlReader;
 
@@ -90,7 +88,7 @@ public class BagConfig {
                 throw new RuntimeException("The " + BAG_INFO_KEY + " key is not present in the " + bagConfigFilePath);
             }
 
-            validateAPTrust(getAPTrustInfo());
+
         } catch (FileNotFoundException e) {
             throw new RuntimeException("The specified bag config file does not exist: " + bagConfigFile
                     .getAbsolutePath());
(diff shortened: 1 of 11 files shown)


escowles:
@ruebot: the [spec](https://tools.ietf.org/html/draft-kunze-bagit-14#section-2.2.2) says: "A metadata element MUST consist of a label, a colon, and a value, each separated by optional whitespace."

So I think it's legal — but it would certainly be easy enough to remove the space before the colons if that looks better.

src/main/java/org/fcrepo/importexport/common/BagConfig.java
```diff
@@ -138,12 +123,34 @@ private void validateAPTrust(final Map<String, String> apTrustInfo) {
...
+     *
+     * @param tagFile name of the tag file to get fields for
      * @return a map of filenames to key-value property maps
      */
-    public Map<String, Map<String, String>> getCustomTags() {
-        final Map<String, Map<String, String>> clone = new HashMap(map);
-        clone.remove(BAG_INFO_KEY);
-        return Collections.unmodifiableMap(clone);
+    public Map<String, String> getFieldsForTagFile(final String tagFile) {
+        if (map.get(tagFile) != null) {
```
escowles: Wouldn't `map.get(tagFile)` return `null` if we just skipped the check?  Is there something else going on here?

src/main/java/org/fcrepo/importexport/common/BagProfile.java
```diff
@@ -111,15 +151,67 @@ public BagProfile(final InputStream in) throws IOException {
...
-     *    or null when no APTrust-Info fields are required.
+     * Get the required tags for the extra tag file
+     *
+     * @param tagFile the tag file to get tags for
+     * @return map of tag = set of acceptable values, or null if tagFile doesn't exist
      */
-    public Map<String, Set<String>> getAPTrustFields() {
-        return aptrustFields;
+    public Map<String, Set<String>> getMetadataFields(final String tagFile) {
+        if (sections.contains(tagFile)) {
```
escowles: Same question as above — I'm not sure why this check is here when the map accessor would already return null if the key didn't exist.

src/main/java/org/fcrepo/importexport/common/BagProfile.java
```diff
@@ -111,15 +151,67 @@ public BagProfile(final InputStream in) throws IOException {
...
+        return null;
+    }
+
+    /**
+     * Does a specific tag section exist in Other-Info
+     *
+     * @param section section to look for
+     * @return true if tagFile exists
+     */
+    public boolean hasSection(final String section) {
```
escowles: This seems to be only used for testing — can't we get enough info about the fields with `getMetadataFields` and `getSectionNames`?

src/main/java/org/fcrepo/importexport/common/BagProfile.java
```diff
@@ -111,15 +151,67 @@ public BagProfile(final InputStream in) throws IOException {
...
+        return (!sections.isEmpty() && sections.contains(section));
+    }
+
+    /**
+     * Does the profile have other tag files in Other-Info
+     *
+     * @return true if any extra tag files exist
+     */
+    public boolean hasExtraSections() {
```
escowles: Same as above — this is only used for testing, can't we just use getSectionNames()?

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -131,29 +131,22 @@ public Exporter(final Config config, final FcrepoClient.FcrepoClientBuilder clie
...
-                    this.bag.addTags(tagfile, customTags.get(tagfile));
+                for (final String tagFile : bagConfig.getTagFiles()) {
+                    this.bag.addTags(tagFile, bagConfig.getFieldsForTagFile(tagFile));
                 }
+
             } catch (NoSuchAlgorithmException e) {
                 // never happens with known algorithm names
-            } catch (Exception e) {
-                throw new RuntimeException("Error loading Bag profile: " + e.toString());
+            } catch (FileNotFoundException e) {
```
escowles: 👍 — I almost made an issue for this — thanks for cleaning up the Exception handling here with much better error messages.

whikloj:
@escowles All good points, I'm not sure what to say about the map accessors. I could claim I was confused by the Jackson get(), but I think I just didn't look. So good catches on them all. 
I also added 2 tests to test for not having an entire file section (ie. aptrust-info.txt) in your config. First that it is okay if the section is not required and then that it throws an Exception if the section has required fields.

escowles:
👍

PR #74, opened 2017-01-17
Adding option to retrieve and export external content

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2382


src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -121,6 +121,13 @@
                 .desc("When present this flag indicates that binaries should be imported/exported.")
                 .required(false).build());
 
+        // Retrieve external content
+        configOptions.addOption(Option.builder("x")
+                .longOpt("external")
+                .hasArg(false)
+                .desc("When present this flag indicates that external content should be retrieved.")
```
awoods: Wording suggesting: `...that external content should be exported`.

PR #76, opened 2017-05-15
Adding option to allow removing tombstones when re-importing over deleted resources

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2202

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -25,7 +25,6 @@ import static org.slf4j.helpers.NOPLogger.NOP_LOGGER;
 
 import java.io.File;
 import java.net.URI;
-import java.text.ParseException;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -157,7 +156,7 @@ public class Config {
     }
 
     /**
-     * Returns true if tombstones should be deleted then importing.
+     * Returns true if tombstones should be deleted when importing.
      *
      * @return retrieve overwite tombstones flag
      */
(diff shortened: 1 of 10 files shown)


src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -386,28 +405,29 @@ public static Config fromFile(final Map<String, String> configVars) throws Parse
...
-                if (entry.getValue().equalsIgnoreCase("true") || entry.getValue().equalsIgnoreCase("false")) {
-                    c.setIncludeBinaries(Boolean.parseBoolean(entry.getValue()));
-                } else {
-                    throw new ParseException(String.format(
-                        "binaries configuration parameter only accepts \"true\" or \"false\", \"{}\" received",
-                        entry.getValue()), lineNumber);
-                }
+                c.setIncludeBinaries(parseBoolean(entry, "binaries", lineNumber));
+            } else if (entry.getKey().trim().equalsIgnoreCase("overwriteTombstones")) {
+                c.setIncludeBinaries(parseBoolean(entry, "overwriteTombstones", lineNumber));
```
bseeger: Shouldn't this be setOverwriteTombstones and not setIncludeBinaries?

src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -386,28 +405,29 @@ public static Config fromFile(final Map<String, String> configVars) throws Parse
...
             } else if (entry.getKey().trim().equalsIgnoreCase("external")) {
-                if (entry.getValue().equalsIgnoreCase("true") || entry.getValue().equalsIgnoreCase("false")) {
-                    c.setRetrieveExternal(Boolean.parseBoolean(entry.getValue()));
-                } else {
-                    throw new ParseException(String.format(
-                        "external configuration parameter only accepts \"true\" or \"false\", \"{}\" received",
-                        entry.getValue()), lineNumber);
-                }
+                c.setIncludeBinaries(parseBoolean(entry, "external", lineNumber));
```
bseeger: Should this be setIncludeBinaries()?   Just checking.  I haven't read all the code around external binaries.

bseeger: Looks like it should be c.setRetrieveExternal, based on the old code.

src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -386,28 +405,29 @@ public static Config fromFile(final Map<String, String> configVars) throws Parse
...
             } else {
                 throw new ParseException(String.format("Unknown configuration key: {}", entry.getKey()), lineNumber);
             }
         }
         return c;
     }
 
+    private static boolean parseBoolean(final Map.Entry<String, String> entry, final String key,
```
awoods: Minor suggestion, the redundant `entry.getValue()` calls in this method could be eliminated by passing in the entry value into this method.

src/main/java/org/fcrepo/importexport/common/Config.java
awoods: These simple logic errors are typically caught in unit tests.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -387,7 +390,27 @@ private FcrepoResponse importContainer(final URI uri, final Model model) throws
             logger.debug("Using Bagit checksum ({}) for file ({})", checksum, containerFile.getPath());
             builder = builder.digest(checksum);
         }
-        return builder.preferLenient().perform();
+        final FcrepoResponse response = builder.preferLenient().perform();
+        if (response.getStatusCode() == 410 && config.overwriteTombstones()) {
+            deleteTombstone(response);
+            return builder.preferLenient().perform();
```
awoods: I am a little surprised that calling the `builder` twice works... particularly given the inputstream (https://github.com/fcrepo4-labs/fcrepo-import-export/pull/76/files#diff-f6498b3791251ba98aa82e1e70ebbe51R384).

It looks like the integration test is working with an empty body: https://github.com/fcrepo4-labs/fcrepo-import-export/pull/76/files#diff-d39f7f254dc82ffc7b294a4db96d2c5eR331

src/main/java/org/fcrepo/importexport/common/Config.java
escowles: It looks like these errors were not caught by unit tests because (1) there was no test and (2) this is dead code duplicating the parsing in ArgParser.  I'll remove this code and add a test for the ArgParser version.

src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -386,28 +405,29 @@ public static Config fromFile(final Map<String, String> configVars) throws Parse
...
+            final int lineNumber) throws ParseException {
+        if (entry.getValue().equalsIgnoreCase("true") || entry.getValue().equalsIgnoreCase("false")) {
+            return Boolean.parseBoolean(entry.getValue());
+        } else {
+            throw new ParseException(String.format(
+                "configuration parameter \"{}\" only accepts \"true\" or \"false\", \"{}\" received",
+                key, entry.getValue()), lineNumber);
```
escowles: Note the extra close paren after `entry.getValue()` — that's the end of the `String.format()` call — the `lineNumber` argument is to the outer `ParseException` constructor.  Passing the value instead of the entry will help slim this down and make that more legible.

src/main/java/org/fcrepo/importexport/importer/Importer.java
escowles: I think this will work if the InputStream supports the `reset` method — which would be true for `ByteArrayInputStream` used and for `FileInputStream` (which is what the client would actually be using).  In theory, I could imagine future work that let you export from one Fedora and import into another in one step (without touching disk), and in that case, we'd need to support non-repeatable InputStreams like you get from HTTP responses.

I will add a test that uses a FileInputStream and has the binary be the resource with the tombstone, to make sure that the FileInputStream is actually repeatable.

escowles: I'm glad I wrote that test, because it didn't work.  So I wound up updating the tests to test the content of both containers and binaries, and found that I needed to create new PutBuilders (not just reset the body).  I'll push a new commit with that update.

bseeger:
👍  for the resolution of the comments I had.  One last question -- why 'T' versus 't'?  Mostly just curious.   Thanks!

escowles:
@bseeger I was calling the option `replaceTombstones` and using `R` because `r` was taken.  But now that I changed the option name and am using `T`, it could just as easily be `t`.  I'll change to that for consistency with the other options.

awoods:
Can you please rebase this on master?

PR #81, opened 2017-05-17
Mapping base paths when importing into a repository with a different base URI

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2335

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -49,6 +49,7 @@ public class Config {
     private String mode;
     private URI resource;
     private URI source;
+    private URI destination;
     private File baseDirectory;
 
     private boolean includeBinaries = false;
@@ -196,41 +197,51 @@ public class Config {
     }
 
     /**
-     * Sets the URI of the source resource, for mapping URIs being imported
+     * Sets the URI map, for mapping URIs being imported
      *
-     * @param source URI to import/export
+     * @param map Comma-separated URIs for the export baseURL and the import baseURL
      */
-    public void setSource(final String source) {
-        if (source != null) {
-            setSource(URI.create(source));
+    public void setMap(final String[] map) {
+        if (map.length == 2 && map[0] != null && map[1] != null) {
+            this.source = URI.create(map[0]);
+            this.destination = URI.create(map[1]);
         }
     }
 
     /**
-     * Sets the URI of the source resoruce, for mapping URIs being imported
+     * Gets the source (export) baseURL, for mapping URIs being imported
      *
-     * @param source URI to import/export
-     */
-    public void setSource(final URI source) {
-        if (source.toString().endsWith("/")) {
-            this.source = URI.create(source.toString().substring(0, source.toString().length() - 1));
-        } else {
-            this.source = source;
-        }
-    }
-
-    /**
-     * Gets the URI of the source resoruce, for mapping URIs being imported
-     *
-     * @return source source
+     * @return Export baseURL
      */
     public URI getSource() {
-        // If 'source' exists, use it... else return 'resource'
-        if (source != null) {
-            return source;
-        } else {
-            return resource;
-        }
+        return source;
+    }
+
+    /**
+     * Gets the source (export) base path, for mapping URIs being imported
+     *
+     * @return Export base path
+     */
+    public String getSourcePath() {
+        return (source == null) ? null : source.getPath();
+    }
+
+    /**
+     * Gets the destination (import) baseURL, for mapping URIs being imported
+     *
+     * @return Import baseURL
+     */
+    public URI getDestination() {
+        return destination;
+    }
+
+    /**
+     * Gets the destination (import) base path, for mapping URIs being imported
+     *
+     * @return Import base path
+     */
+    public String getDestinationPath() {
+        return (destination == null) ? null : destination.getPath();
     }
 
     /**
@@ -385,8 +396,8 @@ public class Config {
         final Map<String, String> map = new HashMap<String, String>();
         map.put("mode", (this.isImport() ? "import" : "export"));
         map.put("resource", this.getResource().toString());
-        if (!this.getSource().toString().isEmpty()) {
-            map.put("source", this.getSource().toString());
+        if (this.getSource() != null && this.getDestination() != null) {
+            map.put("map", this.getSource() + "," + this.getDestination());
         }
         map.put("dir", this.getBaseDirectory().getAbsolutePath());
         if (!this.getRdfLanguage().isEmpty()) {
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -199,12 +199,14 @@ public class Config {
     /**
      * Sets the URI map, for mapping URIs being imported
      *
-     * @param map Comma-separated URIs for the export baseURL and the import baseURL
+     * @param map Array containing two URIs for the export baseURL and the import baseURL
      */
     public void setMap(final String[] map) {
         if (map.length == 2 && map[0] != null && map[1] != null) {
             this.source = URI.create(map[0]);
             this.destination = URI.create(map[1]);
+        } else {
+            throw new IllegalArgumentException("The map should contain the export and import baseURLs");
         }
     }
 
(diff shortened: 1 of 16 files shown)


src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -196,41 +197,51 @@ public URI getResource() {
...
+     * Sets the URI map, for mapping URIs being imported
      *
-     * @param source URI to import/export
+     * @param map Comma-separated URIs for the export baseURL and the import baseURL
      */
-    public void setSource(final String source) {
-        if (source != null) {
-            setSource(URI.create(source));
+    public void setMap(final String[] map) {
+        if (map.length == 2 && map[0] != null && map[1] != null) {
```
awoods: Should we throw an `IllegalArgumentException` in an `else` block?

src/main/java/org/fcrepo/importexport/common/TransferProcess.java
```diff
@@ -81,49 +81,75 @@ public static String decodePath(final String encoded) {
...
+        return new File(baseDir, encodePath(relativePath(uri, sourcePath, destinationPath)) + extension);
+    }
+
+    /**
+     * Translates the path of a URI relative to an optional source URI.
+     * @param uri the URI for the resource
+     * @param sourcePath the base path the resource was exported from
+     * @param destinationPath the base path the resource is importing to
+     */
+    public static String relativePath(final URI uri, final String sourcePath, final String destinationPath) {
```
awoods: The `TransferProcess` interface is used by both the `Importer` and `Exporter`... however, this method appears to be written in the context of the `Exporter`... whereas the analogous `Importer` capability is found here:
https://github.com/fcrepo4-labs/fcrepo-import-export/pull/81/files#diff-f6498b3791251ba98aa82e1e70ebbe51R512

Would it be more clear to name this method differently?

src/main/java/org/fcrepo/importexport/common/TransferProcess.java
```diff
@@ -81,49 +81,75 @@ public static String decodePath(final String encoded) {
...
      */
-    public static File fileForURI(final URI uri, final File baseDir, final String extension) {
-        return new File(baseDir, TransferProcess.encodePath(uri.getPath()) + extension);
+    public static File fileForURI(final URI uri, final String sourcePath, final String destinationPath,
+            final File baseDir, final String extension) {
+        return new File(baseDir, encodePath(relativePath(uri, sourcePath, destinationPath)) + extension);
+    }
+
+    /**
+     * Translates the path of a URI relative to an optional source URI.
```
awoods: Do you mean `optional source path` instead of `optional source URI`?

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    public void testMappingNull() throws Exception {
+        final URI uri = create("http://localhost:8080/rest/foo");
+        assertEquals(new File(dir, "rest/foo" + ext), fileForURI(uri, null, null, dir, ext));
+    }
+
+    @Test
+    public void testMappingPort() throws Exception {
+        final URI src = create("http://localhost:9090/rest/foo");
```
awoods: `src` not used.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final URI uri = create("http://localhost:8080/rest/foo");
+        assertEquals(new File(dir, "rest/foo" + ext), fileForURI(uri, "/rest", "/rest", dir, ext));
+    }
+
+    @Test
+    public void testMappingHost() throws Exception {
+        final URI src = create("http://example.org:8080/rest/foo");
```
awoods: `src` not used.
..therefore, this test is not testing what the name implies: `testMappingHost()`

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final URI uri = create("http://localhost:8080/rest/foo");
+        assertEquals(new File(dir, "rest/foo" + ext), fileForURI(uri, "/rest", "/rest", dir, ext));
+    }
+
+    @Test
+    public void testMappingBase() throws Exception {
+        final URI src = create("http://localhost:9090/rest/foo");
```
awoods: `src` not used.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final URI uri = create("http://localhost:8080/fedora/rest/foo");
+        assertEquals(new File(dir, "rest/foo" + ext), fileForURI(uri, "/rest", "/fedora/rest", dir, ext));
+    }
+
+    @Test
+    public void testMappingPath() throws Exception {
+        final URI src = create("http://localhost:7777/rest/dev/foo");
```
awoods: `src` not used.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final URI uri = create("http://localhost:8888/rest/prod/foo");
+        assertEquals(new File(dir, "rest/dev/foo" + ext), fileForURI(uri, "/rest/dev", "/rest/prod", dir, ext));
+    }
+
+    @Test
+    public void testMappingRest() throws Exception {
+        final URI src = create("http://localhost:9090/rest/bar");
```
awoods: `src` not used.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final URI uri = create("http://localhost:8080/rest/foo");
+        assertEquals(new File(dir, "rest/foo" + ext), fileForURI(uri, "/rest", "/rest", dir, ext));
+    }
+
+    @Test
+    public void testMappingRestless() throws Exception {
+        final URI src = create("http://localhost:9090/foo");
```
awoods: `src` not used.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final URI uri = create("http://localhost:8080/rest/foo");
+        assertEquals(new File(dir, "foo" + ext), fileForURI(uri, "/", "/rest/", dir, ext));
+    }
+
+    @Test
+    public void testMappingRestless2() throws Exception {
+        final URI src = create("http://localhost:9090/rest/foo");
```
awoods: `src` not used.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.io.File;
+import java.net.URI;
+
+import org.junit.Test;
+
+/**
+ * @author escowles
+ * @since 2016-12-14
+ */
+public class TransferProcessTest {
```
awoods: Since `src` is not used in any of the tests below... assumptions around what is being tested are not always correct.

src/main/java/org/fcrepo/importexport/common/TransferProcess.java
escowles: `relativePath` isn't used directly by Importer or Exporter, but both of those use `fileForURI` which uses it.  So it should probably be private (or just folded into fileForURI).

awoods: Either `private` or rolled into `fileForURI` sound good.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
escowles: I created the unused `src` values when brainstorming the possible permutations of changing host/port/path/etc.  But since they are not being used in the tests, I can remove them.

src/test/java/org/fcrepo/importexport/common/TransferProcessTest.java
escowles: Yes, host and port are irrelevant in these tests because we're only using the path.  I'll remove these tests.

PR #82, opened 2017-05-18
Added support for round-tripping with binaries excluded.

lsitu:
Related to https://jira.duraspace.org/browse/FCREPO-2426.

Supporting round-tripping with binaries excluded by not exporting references to the binary resources.

### DIFF
--- a/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
+++ b/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
@@ -24,10 +24,12 @@ import static org.apache.jena.rdf.model.ResourceFactory.createProperty;
 import static org.apache.jena.riot.RDFLanguages.contentTypeToLang;
 import static org.fcrepo.importexport.common.FcrepoConstants.CONTAINER;
 import static org.fcrepo.importexport.common.FcrepoConstants.NON_RDF_SOURCE;
+import static org.fcrepo.importexport.common.FcrepoConstants.REPOSITORY_NAMESPACE;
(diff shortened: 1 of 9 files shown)


src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -303,14 +333,61 @@ private void exportMembers(final File file) {
...
+        return model;
+    }
+
+    /**
+     * Method to find and set the repository root from the resource uri.
+     * @param uri the URI for the resource
+     * @throws IOException
+     * @throws FcrepoOperationFailedException
+     */
+    private void findRepositoryRoot(final URI uri) throws IOException, FcrepoOperationFailedException {
```
awoods: Can we consolidate this method and:
https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/importer/Importer.java#L651

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -303,14 +333,61 @@ private void exportMembers(final File file) {
             logger.warn("Unable to parse file: {}", ex.toString());
         }
     }
-    void writeResponse(final FcrepoResponse response, final List<URI> describedby, final File file)
+
+    /**
+     * Filter out the binary resource references from the model
+     * @param model
```
awoods: Please add description to the `param` and `return`

PR #84, opened 2017-05-22
Adding support for exporting inbound references

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2453

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,10 @@ into `http://example.org:8080/fedora/rest/`:
 java -jar target/fcrepo-import-export-0.1.0-SNAPSHOT.jar --mode import --resource http://example.org:8080/fedora/rest/ --dir /tmp/test --binaries --map http://localhost:8984/rest/dev/,http://example.org:8080/fedora/rest/
 ```
 
+To retrieve inbound references (for example, when exporting a collection and you also want to export the members that link to the collection), use the `-i` or `--inbound` option.  When enabled, resources that are linked to or from with the specified predicates will be exported.
+
+To retrieve external content binaries (binaries on other systems linked to with the `message/external-body` content type), use the `-x` or `--external` option.  When enabled, the external binaries will be retrieved and included in the export.  When disabled, they will not be retrieved, and only pointers to them will be exported.
+
 Running the import/export utility with a BagIt support
 ------------------------------------------------------
 
@@ -159,6 +163,8 @@ That configuration file is [Yaml](http://yaml.org) and allows for the following 
 * dir: Directory to import from/export to
 * binaries: [true|false] # whether is import/export binary resources
 * overwriteTombstones: [true|false] # whether to replace tombstones of previously deleted resources
+* external: [true|false] # whether to retrieve external content binaries when exporting
+* inbound: [true|false] # whether to export inbound references when exporting
 * map: Old and new base URIs, separated by comma, to map URIs when importing.
 * resource: The resource to export/import
 * rdfLang: The RDF language to export into or import from
--- a/src/main/java/org/fcrepo/importexport/ArgParser.java
+++ b/src/main/java/org/fcrepo/importexport/ArgParser.java
@@ -128,6 +128,13 @@ public class ArgParser {
                 .desc("When present this flag indicates that external content should be exported.")
                 .required(false).build());
 
+        // Retrieve inbound references
+        configOptions.addOption(Option.builder("i")
+                .longOpt("inbound")
+                .hasArg(false)
+                .desc("When present this flag indicates that inbound references should be exported.")
+                .required(false).build());
+
         // Overwrite Tombstones
         configOptions.addOption(Option.builder("t")
                  .longOpt("overwriteTombstones")
@@ -322,6 +329,7 @@ public class ArgParser {
         config.setBaseDirectory(cmd.getOptionValue('d'));
         config.setIncludeBinaries(cmd.hasOption('b'));
         config.setRetrieveExternal(cmd.hasOption('x'));
+        config.setRetrieveInbound(cmd.hasOption('i'));
         config.setOverwriteTombstones(cmd.hasOption('t'));
 
         final String rdfLanguage = cmd.getOptionValue('l');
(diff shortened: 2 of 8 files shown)


awoods:
@escowles : please rebase on master

src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -487,6 +495,8 @@ public static Config configFromFile(final Map<String, String> configVars) throws
                 c.setIncludeBinaries(parseBoolean("binaries", entry.getValue(), lineNumber));
             } else if (entry.getKey().trim().equalsIgnoreCase("external")) {
                 c.setRetrieveExternal(parseBoolean("external", entry.getValue(), lineNumber));
+            } else if (entry.getKey().trim().equalsIgnoreCase("inbound")) {
+                c.setRetrieveInbound(parseBoolean("external", entry.getValue(), lineNumber));
```
awoods: typo: `external` should be `inbound`

src/main/java/org/fcrepo/importexport/common/FcrepoConstants.java
```diff
@@ -59,4 +59,7 @@
     public static final Resource PAIRTREE = createResource(REPOSITORY_NAMESPACE + "Pairtree");
 
     public static final String BAG_INFO_FIELDNAME = "Bag-Info";
+
+    public static final Resource INBOUND_REFERENCES =
+            createResource("http://fedora.info/definitions/v4/repository#InboundReferences");
```
awoods: Please use the constant value: `REPOSITORY_NAMESPACE`

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -276,9 +281,17 @@ private void exportDescription(final URI uri) throws FcrepoOperationFailedExcept
...
+        } else if (file.exists()) {
+            logger.info("Already exported {}", uri);
+            return;
         }
 
-        try (FcrepoResponse response = client().get(uri).accept(config.getRdfLanguage()).perform()) {
+        GetBuilder getBuilder = client().get(uri).accept(config.getRdfLanguage());
+        if (config.retrieveInbound()) {
+            getBuilder = getBuilder.preferRepresentation(
+                Arrays.asList(new URI[]{ URI.create(INBOUND_REFERENCES.getURI()) }), null);
```
awoods: You can remove the `new URI[] {...}` to:
```
Arrays.asList(URI.create(INBOUND_REFERENCES.getURI())), null);
```

README.md
```diff
@@ -91,6 +91,10 @@ into `http://example.org:8080/fedora/rest/`:
 java -jar target/fcrepo-import-export-0.1.0-SNAPSHOT.jar --mode import --resource http://example.org:8080/fedora/rest/ --dir /tmp/test --binaries --map http://localhost:8984/rest/dev/,http://example.org:8080/fedora/rest/
 ```
 
+To retrieve inbound references (for example, when exporting a collection and you also want to export the members that link to the collection), use the `-i` or `--inbound` option.  When enabled, resources that are linked to or from with the specified predicates will be exported.
+
+To retrieve external content binaries (binaries on other systems linked to with the `message/external-body` content type), use the `-x` or `--external` option.  When enabled, the external binaries will be retrieved and included in the export.  When disable, they will not be retrieved, and only pointers to them will be exported.
```
awoods: Typo: `When disable` should be `When disabled`

README.md
```diff
@@ -159,6 +163,8 @@ That configuration file is [Yaml](http://yaml.org) and allows for the following
 * dir: Directory to import from/export to
 * binaries: [true|false] # whether is import/export binary resources
 * overwriteTombstones: [true|false] # whether to replace tombstones of previously deleted resources
+* external: [true|false] # whether to retrieve external content binaries when export
```
awoods: Typo: `when export` should be `when exporting`

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -289,14 +302,23 @@ private void exportDescription(final URI uri) throws FcrepoOperationFailedExcept
             exportLogger.error(String.format("Error exporting description: {}, Cause: {}", uri, ex.getMessage()), ex);
         }
 
-        exportMembers(file);
+        exportMembers(file, uri);
     }
```
awoods: This is a minor stylistic point, and if you have a specific personal preference for omitting whitespace after the `}` at the end of this method, I would be interested to hear it.
Otherwise, for legibility, please add a newline between these two methods.

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -289,14 +302,23 @@ private void exportDescription(final URI uri) throws FcrepoOperationFailedExcept
...
-    private void exportMembers(final File file) {
+    private void exportMembers(final File file, final URI parentURI) {
```
awoods: Same note as above: please add a new line after the `exportMembers` method.

src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -130,6 +130,13 @@
                 .desc("When present this flag indicates that external content should be exported.")
                 .required(false).build());
 
+        // Retrieve inbound references
+        configOptions.addOption(Option.builder("i")
```
awoods: What is the expected behavior when the `-i` option is provided during an `import`?

escowles: This should not have any impact when importing — if the inbound references were exported, then they'll be found on disk and imported, if they weren't exported then they won't.

There are a couple of options that only make sense in either import or export contexts — with the general ability to reuse commands between import/export (and particularly with the config file), I don't think we should refuse to accept the options for the operations where they won't have any effect.  But maybe we should document this more clearly?  Maybe the CLI parser has some notion of option groups that would improve the help output, etc.?

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
escowles: OK, I've added a blank line — I agree that it's nice to have a blank line between methods, so I'm not sure why there wasn't one.

src/main/java/org/fcrepo/importexport/ArgParser.java
awoods: I can see value in making the use of various options clear with respect to their applicability to either import or export or both.
That could be done in the CLI help output and/or with WARNING log output with unused arguments.

awoods:
@escowles : with @lsitu's recent update... another rebase on master is required. For what it's worth, I do not plan on merging any other commits until this one is in.

escowles:
@awoods No problem — the rebase was pretty easy both times!

PR #85, opened 2017-05-22
FCREPO-2458 Export versions

bbpennel:
Adds configurable option for export of versioned resources

https://jira.duraspace.org/browse/FCREPO-2458

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -478,6 +478,7 @@ public class Config {
         map.put("inbound", Boolean.toString(this.retrieveInbound));
         map.put("overwriteTombstones", Boolean.toString(this.overwriteTombstones()));
         map.put("legacyMode", Boolean.toString(this.isLegacy()));
+        map.put("versions", Boolean.toString(this.includeVersions));
         if (this.getBagProfile() != null) {
             map.put("bag-profile", this.getBagProfile());
         }
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -59,6 +59,7 @@ public class Config {
     private boolean retrieveInbound = false;
     private boolean overwriteTombstones = false;
     private boolean legacy = false;
+    private boolean includeVersions = false;
     private String bagProfile = null;
     private String bagConfigPath = null;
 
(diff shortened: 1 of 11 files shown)


src/test/java/org/fcrepo/importexport/integration/AbstractResourceIT.java
```diff
@@ -50,7 +50,7 @@
  * @author awoods
  * @since 2016-09-01
  */
-abstract class AbstractResourceIT {
+public abstract class AbstractResourceIT {
```
bbpennel: I couldn't execute the ITs in my IDE without this change, although they ran via commandline

src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -188,6 +189,24 @@ public boolean overwriteTombstones() {
     }
 
     /**
```
awoods: You should add `includeVersions` to [getMap()](https://github.com/bbpennel/fcrepo-import-export/blob/3e82757cfd76adedbfa9d0a919a30859bf01104c/src/main/java/org/fcrepo/importexport/common/Config.java#L465)

..and update the following two tests as well:
- [testSerializedConfigCustom](https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/test/java/org/fcrepo/importexport/ArgParserTest.java#L319)
- [testSerializedConfigDefault](https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/test/java/org/fcrepo/importexport/ArgParserTest.java#L346)

src/test/java/org/fcrepo/importexport/exporter/ExportVersionsTest.java
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ * 
+ * @author bbpennel
+ *
+ */
+public class ExportVersionsTest {
+
+    private final String BASE_URI = "http://localhost:8080/rest/";
+    private String exportDirectory = "target/export";
```
awoods: For clarity, maybe change this to: `target/export-versions-test`

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -244,6 +247,8 @@ private void export(final URI uri) {
                 exportBinary(uri, describedby, external);
             } else if (linkHeaders.contains(containerURI)) {
                 exportDescription(uri);
+                // Export versions for this container
+                exportVersions(uri);
```
bbpennel: Yes, this was somewhat unfortunate but intentional.  For binaries `exportVersions` has to be called after a check of `config.isIncludeBinaries`, which happens inside of `exportBinary`.  But I couldn't mirror that arrangment with containers since `exportDescription` is called by binaries as well (having it in there caused issues with exporting the wrong URI for binaries).  

I could have moved the `isIncludeBinaries` check out of exportBinary, but that seemed out of line with how other methods were doing precondition checks.  If you'd prefer I can move it though.

awoods: No, your rationale is reasonable. I just wanted to make sure there was a reason.

PR #89, opened 2017-05-24
Add file describing import-export format.  Addresses one portion of F…

coblej:
…CREPO-2455.

### DIFF
--- a/import-export-format.md
+++ b/import-export-format.md
@@ -3,17 +3,17 @@ Import / Export Format
 
 Export Format
 -------------
-When run in `export` mode, the import/export utility creates a directory tree representing the designated resource node and its child nodes:
-* The directory structure represents the hierarchy of nodes beginning at the Fedora repository root down to the exported resource.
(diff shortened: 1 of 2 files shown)


import-export-format.md
```diff
@@ -0,0 +1,183 @@
+Import / Export Format
+======================
+
+Export Format
+-------------
+When run in `export` mode, the import/export utility creates a directory tree representing the designated resource node and its child nodes:
```
awoods: Please change: `designated resource node` 
..to `designated resource`
- and `child nodes` 
..to `child resources (where 'b' is a child of 'a' in this case, /a/b)`

import-export-format.md
```diff
@@ -0,0 +1,183 @@
...
+* The directory structure represents the hierarchy of nodes beginning at the Fedora repository root down to the exported resource.
```
awoods: In general, please perform a find/replace of `node` in favor of `resource`.

PR #92, opened 2017-05-24
Making findRepositoryRoot more robust

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2477

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,12 +89,6 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>22.0</version>
-    </dependency>
-
 
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,12 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>22.0</version>
+    </dependency>
+
 
   </dependencies>
 
--- a/src/main/java/org/fcrepo/importexport/importer/Importer.java
+++ b/src/main/java/org/fcrepo/importexport/importer/Importer.java
@@ -167,7 +167,8 @@ public class Importer implements TransferProcess {
     public void run() {
         logger.info("Running importer...");
 
-        findRepositoryRoot(config.getResource());
+        repositoryRoot = findRepositoryRoot(config.getResource());
+        logger.debug("Repository root {}", repositoryRoot);
 
         processImport(config.getResource());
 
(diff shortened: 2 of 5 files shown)


src/test/java/org/fcrepo/importexport/integration/ImporterIT.java
```diff
@@ -279,6 +279,33 @@ public void testImportMappedTrailingSlash() throws Exception {
         assertTrue(resourceExists(childURI));
     }
 
+    @Test
+    public void testRepositoryRoot() throws Exception {
```
awoods: These seem more like unit tests than integration tests. Would you mind moving them to ImporterTest.java?

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -59,6 +59,8 @@
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
 
+import com.google.common.annotations.VisibleForTesting;
```
awoods: If we are only bringing in `guava` for this annotation, I would suggest removing the dependency and simply adding a method comment.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -648,26 +651,27 @@ public static boolean verifyBag(final File bagDir) {
     /**
      * Method to find and set the repository root from the resource uri.
      * @param uri the URI for the resource
-     * @throws IOException
-     * @throws FcrepoOperationFailedException
      */
-    private void findRepositoryRoot(final URI uri) {
-        repositoryRoot = uri;
+    @VisibleForTesting
+    public URI findRepositoryRoot(final URI uri) {
```
awoods: You can likely make the `protected` if the unit test is moved to `ImporterTest.java`.

src/test/java/org/fcrepo/importexport/integration/ImporterIT.java
escowles: For the most part, I'd say that's true — I've pushed a commit to move the second test to the unit test, and added a few others handling the main cases with mocks.  But I'd also like to keep the one test that verifies that we are finding the correct root of the repository we're running our integration tests against.

src/main/java/org/fcrepo/importexport/importer/Importer.java
escowles: If we keep the one integration test, this will need to continue to be public.  Since this is primarily a command-line utility, and the method doesn't impact the internal state of the importer instance, I think this is fine.

awoods: The following test does not seem like an integration test (i.e. it is not actually making any HTTP requests):
https://github.com/fcrepo4-labs/fcrepo-import-export/blob/0fab8b4e9b850605e557d42ac3007563760396d9/src/test/java/org/fcrepo/importexport/integration/ImporterIT.java#L283

But if you want to keep it in `ImporterIT.java`, I can respect that.
Let me know either way, and we can either move forward with the code as it stands now, or with another commit that moves the mentioned test method to `ImporterTest.java` and reduces access on this method to `protected`.

escowles: I believe all of those `findRepositoryRoot` calls do make HTTP requests.  some of them will get 404 responses, but the `findRepositoryRoot` logic will continue chopping off path elements until it finds the root resource.

pom.xml
```diff
@@ -89,6 +89,12 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
awoods: It looks like `guava` is still here. I am happy to remove this dependency and move this PR through... unless you would prefer to add a commit.

escowles: I've pushed a new commit to remove Guava

awoods:
Please resolve merge conflict.

PR #93, opened 2017-05-25
Filtering out inbound links from serialized RDF

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2483 and https://jira.duraspace.org/browse/FCREPO-2481


awoods:
It looks like this PR needs a rebase on master.

PR #96, opened 2017-05-25
Only write config files if the --writeConfig option is specified

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2430

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -53,6 +53,7 @@ public class Config {
     private URI source;
     private URI destination;
     private File baseDirectory;
+    private File writeConfig;
 
     private boolean includeBinaries = false;
     private boolean retrieveExternal = false;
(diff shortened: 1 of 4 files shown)


awoods:
The following test fails on Travis and for me locally if I remove the old `/tmp/importexport.yml` file.
https://travis-ci.org/fcrepo4-labs/fcrepo-import-export/builds/236155126#L3283

We should probably remove: https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/ArgParser.java#L69 ...and everywhere it is used.

src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -134,6 +135,22 @@ public boolean isIncludeBinaries() {
     }
 
     /**
+     * Sets the location to write a sample config file.
```
awoods: We should probably include this arg in the config file, itself, no?
https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/common/Config.java#L465

...as well as the tests:
- https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/test/java/org/fcrepo/importexport/ArgParserTest.java#L329
- https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/test/java/org/fcrepo/importexport/ArgParserTest.java#L357

escowles: I don't think we should include the `writeConfig` option in the config file.  The option triggers writing the file to disk, but I wouldn't want subsequent uses of the config file to overwrite itself (which is what including the option in the file itself would trigger).  If everything worked properly, then the config file wouldn't change.  If anything went amiss, it seems like you could mess up the config you were using.

PR #99, opened 2017-05-26
Filtering all inbound references, regardless of predicate

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2481 and https://jira.duraspace.org/browse/FCREPO-2500

### DIFF
--- a/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
+++ b/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
@@ -33,6 +33,8 @@ import static org.fcrepo.importexport.common.TransferProcess.fileForBinary;
 import static org.fcrepo.importexport.common.TransferProcess.fileForExternalResources;
 import static org.fcrepo.importexport.common.TransferProcess.fileForURI;
 import static org.fcrepo.importexport.common.TransferProcess.isRepositoryRoot;
+import static org.fcrepo.importexport.common.UriUtils.withSlash;
+import static org.fcrepo.importexport.common.UriUtils.withoutSlash;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.io.ByteArrayInputStream;
@@ -317,7 +319,7 @@ public class Exporter implements TransferProcess {
             final String responseBody = IOUtils.toString(response.getBody());
             final Model model = createDefaultModel().read(new ByteArrayInputStream(responseBody.getBytes()),
                     null, config.getRdfLanguage());
-            List<URI> inboundMembers = null;
+            Set<URI> inboundMembers = null;
 
             if (!config.isIncludeBinaries() || config.retrieveInbound()) {
 
@@ -353,18 +355,17 @@ public class Exporter implements TransferProcess {
 
     }
 
-    private List<URI> filterInboundReferences(final URI uri, final Model model) {
-        final List<URI> inboundMembers = new ArrayList<>();
+    private Set<URI> filterInboundReferences(final URI uri, final Model model) {
+        final String withSlash = withSlash(uri).toString();
+        final String withoutSlash = withoutSlash(uri).toString();
+        final Set<URI> inboundMembers = new HashSet<>();
         final List<Statement> removeList = new ArrayList<>();
-        for (final String p : config.getPredicates()) {
-            final StmtIterator inbound = model.listStatements(null, createProperty(p), (RDFNode)null);
-            while (inbound.hasNext()) {
-                final Statement s = inbound.next();
-                final String subject = s.getSubject().toString();
-                if (!subject.startsWith(uri.toString())) {
-                    removeList.add(s);
-                    inboundMembers.add(URI.create(subject));
-                }
+        for (final StmtIterator inbound = model.listStatements(); inbound.hasNext(); ) {
+            final Statement s = inbound.next();
+            final String subject = s.getSubject().toString();
+            if (!subject.equals(withSlash) && !subject.equals(withoutSlash)) {
+                removeList.add(s);
+                inboundMembers.add(URI.create(subject));
             }
         }
 
@@ -372,7 +373,7 @@ public class Exporter implements TransferProcess {
         return inboundMembers;
     }
 
-    private void exportMembers(final Model model, final List<URI> inboundMembers) {
+    private void exportMembers(final Model model, final Set<URI> inboundMembers) {
         for (final String p : config.getPredicates()) {
             final NodeIterator members = model.listObjectsOfProperty(createProperty(p));
             while (members.hasNext()) {
--- a/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
+++ b/src/main/java/org/fcrepo/importexport/exporter/Exporter.java
@@ -250,7 +250,7 @@ public class Exporter implements TransferProcess {
                 final List<URI> describedby = response.getLinkHeaders("describedby");
                 exportBinary(uri, describedby, external);
             } else if (linkHeaders.contains(containerURI)) {
-                exportDescription(uri);
+                exportDescription(uri, null);
                 // Export versions for this container
                 exportVersions(uri);
             } else {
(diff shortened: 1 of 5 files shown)


escowles:
Note: this branch results in all binary description properties being filtered out: https://jira.duraspace.org/browse/FCREPO-2500

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -353,26 +355,25 @@ private void exportDescription(final URI uri) throws FcrepoOperationFailedExcept
...
-                if (!subject.startsWith(uri.toString())) {
-                    removeList.add(s);
-                    inboundMembers.add(URI.create(subject));
-                }
+        for (final StmtIterator inbound = model.listStatements(); inbound.hasNext(); ) {
+            final Statement s = inbound.next();
+            final String subject = s.getSubject().toString();
+            if (!subject.equals(withSlash) && !subject.equals(withoutSlash)) {
+                removeList.add(s);
+                inboundMembers.add(URI.create(subject));
```
lsitu: @escowles Maybe we need a global inboundMembers list to keep track of those inbound resources that are already exported previously?

escowles: @lsitu I think this change will fix that issue: using a `Set` instead of a `List` so we don't export the same resources over and over: https://github.com/fcrepo4-labs/fcrepo-import-export/pull/99/files#diff-230835395ae42488584f0706d8828ad4R323

src/test/java/org/fcrepo/importexport/common/UriUtilsTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test;
+
+/**
+ * @author escowles
+ * @since 2017-05-26
+ */
+public class UriUtilsTest {
+
+    private final URI withSlash = URI.create("http://localhost:8080/rest/");
+    private final URI withOutSlash = URI.create("http://localhost:8080/rest/");
```
awoods: It looks like this URI should be: `http://localhost:8080/rest` <- note, no trailing slash.

src/test/java/org/fcrepo/importexport/common/UriUtilsTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Test
+    public void testWithSlash() throws Exception {
+        assertEquals(withSlash, UriUtils.withSlash(withSlash));
+        assertEquals(withSlash, UriUtils.withSlash(withOutSlash));
+    }
+
+    @Test
+    public void testWithoutSlash() throws Exception {
+        assertEquals(withOutSlash, UriUtils.withSlash(withSlash));
```
awoods: I assume this test should be invoking the method: `UriUtils.withoutSlash()`

src/test/java/org/fcrepo/importexport/integration/ExporterIT.java
```diff
@@ -317,6 +321,44 @@ public void testExportCustomPredicatesInbound() throws Exception {
...
                 createProperty("http://example.org/custom"),
                 createResource(col1.toString())));
+        assertFalse(collectionModel.contains(
+                createResource(other.toString()),
+                createProperty("http://example.org/other"),
+                createResource(col1.toString())));
+    }
+
+    @Test
+    public void testExportInboundFiles() throws Exception {
```
awoods: This test does not seem to be demonstrating a fix to broken behavior. 
When I checkout the `master` branch and insert nothing from this PR except this test, the build succeeds.
I think the test should be linking a resourceX from outside of the target resourceT hierarchy to test that:
1. The resourceX is exported
2. The model of resourceT does not include mention of resourceX

src/test/java/org/fcrepo/importexport/integration/ExporterIT.java
```diff
@@ -317,6 +321,44 @@ public void testExportCustomPredicatesInbound() throws Exception {
...
+        // export with inbound references
+        final Config config = new Config();
+        config.setMode("export");
+        config.setBaseDirectory(TARGET_DIR + "/" + uuid);
+        config.setResource(baseURI);
+        config.setRdfExtension(DEFAULT_RDF_EXT);
+        config.setRdfLanguage(DEFAULT_RDF_LANG);
+        config.setUsername(USERNAME);
+        config.setPassword(PASSWORD);
+        config.setPredicates(new String[]{ CONTAINS.toString() });
```
awoods: This is the [default predicate](https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/common/Config.java#L49)... and can effectively be omitted.

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
lsitu: @escowles I am may not understand the inbound resources well but is there a case that other resource descriptions may include the same inbound resource reference or not? That is, more than one resource may contains the same inbound reference and when the method exportDescription(final URI uri, final URI binaryURI) is called repeatedly that inbound resource will be processed again and again.

escowles: Yes, if there are many children of a single node, then they might all call `exportDescription` — but there is a pretty quick check at the beginning to avoid re-fetching: https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/exporter/Exporter.java#L302

I think it could be worth checking whether managing a global list of URIs that have already been exported, but we'd need to make sure that didn't get too big or it could be a memory problem for larger exports.

src/test/java/org/fcrepo/importexport/integration/ExporterIT.java
escowles: @awoods This test demonstrates a fix to the bug in d34b5771ff4abeb6ecc52fa1fa528b0117ab1b9f — that I wrote up in [fcrepo-2500](https://jira.duraspace.org/browse/FCREPO-2500) — namely that the first pass at fixing inbound references was stripping all binary description properties (I was seeing a turtle file with only namespace declarations in it).

Would it be helpful to move this commit to its own commit?

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
lsitu: I agree. Thanks.

src/test/java/org/fcrepo/importexport/integration/ExporterIT.java
awoods: Since the test is less related to the rest of the PR, pulling into a separate branch/PR targeting [fcrepo-2500](https://jira.duraspace.org/browse/FCREPO-2500) makes sense.

What are your thoughts on an integration test (potentially as described above) exercising the fix in this current PR?

escowles: @awoods The previous test already demonstrates what you outlined:
* `obj1` is the resource from outside `col1`
* I'm verifying that it is exported at [line 313](https://github.com/fcrepo4-labs/fcrepo-import-export/pull/99/commits/c79a83d135d578ed7ee23c6859a5fe6b8f593493#diff-2085f9c938964a6c935f6979bfc109dbR313)
* I'm verifying that the inbound triple (obj1 ex:custom col1) isn't exported in col1's export on [line 320](https://github.com/fcrepo4-labs/fcrepo-import-export/pull/99/commits/c79a83d135d578ed7ee23c6859a5fe6b8f593493#diff-2085f9c938964a6c935f6979bfc109dbR320)
* And I'm verifying that an inbound triple with a different predicate isn't included in the export on [line 324](https://github.com/fcrepo4-labs/fcrepo-import-export/pull/99/commits/c79a83d135d578ed7ee23c6859a5fe6b8f593493#diff-2085f9c938964a6c935f6979bfc109dbR324)

What I'm adding in c79a83d135d578ed7ee23c6859a5fe6b8f593493 is a test that demonstrates fcrepo-2500 and the fix for it.  I don't think that can be separated from this PR, because d34b5771ff4abeb6ecc52fa1fa528b0117ab1b9f results in all binary descriptions being exported as empty turtle documents (just the namespace definitions survive the filtering).

escowles:
Closing in favor of #102

PR #101, opened 2017-05-27
Preventing export of fcr:versions on repository root

bbpennel:


src/main/java/org/fcrepo/importexport/common/FcrepoConstants.java
```diff
@@ -58,6 +58,7 @@
     public static final String REPOSITORY_NAMESPACE = "http://fedora.info/definitions/v4/repository#";
     public static final Resource INBOUND_REFERENCES = createResource(REPOSITORY_NAMESPACE + "InboundReferences");
     public static final Resource PAIRTREE = createResource(REPOSITORY_NAMESPACE + "Pairtree");
+    public static final Resource REPOSITORY_ROOT = createResource(REPOSITORY_NAMESPACE + "RepositoryRoot");
```
awoods: This new variable is only used in tests... is it necessary?

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2503

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -68,7 +68,6 @@
 import org.apache.jena.rdf.model.Statement;
 import org.apache.jena.rdf.model.StmtIterator;
 import org.apache.jena.riot.RDFDataMgr;
-
```
awoods: Needless formatting changes like this are often a source of merge conflicts. Please remove.

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -81,7 +80,6 @@
 import org.fcrepo.importexport.common.ProfileValidationUtil;
 import org.fcrepo.importexport.common.ResourceNotFoundRuntimeException;
 import org.fcrepo.importexport.common.TransferProcess;
-
```
awoods: Please revert unnecessary removal of newline.

src/test/java/org/fcrepo/importexport/exporter/ExporterTest.java
```diff
@@ -370,53 +373,6 @@ public void testExportBagCustomTags() throws Exception, FcrepoOperationFailedExc
         Assert.assertTrue(customLines.contains("Foo : Bar"));
         Assert.assertTrue(customLines.contains("Baz : Quux"));
     }
-
-    @Test
-    public void testExportVersionsContainers() throws Exception {
```
awoods: Although it looks like the removal of this test method is associated with the addition of:
https://github.com/fcrepo4-labs/fcrepo-import-export/pull/101/files#diff-3dc91ed69dd47e72f6396a8309355095R296

However, they are different. Can you speak to why this one needs to go away?

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
bbpennel: Eclipse strikes again

src/test/java/org/fcrepo/importexport/exporter/ExporterTest.java
bbpennel: The test you referenced was copy and pasted from this method, then cleaned up to use the helper methods provided in ExporterVersionTest and match the style of the other test methods.  I had written ExporterVersionTest to contain all the versioning tests.  Both define the same resources and same assertions, I just forgot to remove this one from its original home when doing refactoring.

src/main/java/org/fcrepo/importexport/common/FcrepoConstants.java
bbpennel: Sorry, I forgot to include it in the non-test method where it was referenced (isRepositoryRoot), it is used now.

PR #105, opened 2017-06-06
Ensure RDF is written to disk in RDF 1.0 syntax

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2515


escowles:
@awoods The tests pass for me locally, but they failed twice in Travis.  Any ideas about why this is failing in Travis?

PR #107, opened 2017-06-22
Updated importer to add contained resources before storing container …

mikedurbin:
…RDF.

https://jira.duraspace.org/browse/FCREPO-2518

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -261,19 +261,30 @@ public class Config {
      */
     public void setMap(final String[] map) {
         if (map.length == 2 && map[0] != null && map[1] != null) {
-            this.source = URI.create(map[0]);
-            this.destination = URI.create(map[1]);
-            checkTrailingSlashes(this.source.toString(), this.destination.toString());
+            final String sourceStr = map[0];
+            final String destinationStr = map[1];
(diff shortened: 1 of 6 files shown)


mikedurbin:
Thanks for testing.  I think I fixed the error associated with binary children.

dbernstein:
Cool I will try it again.

mikedurbin:
This is easy enough to fix, but it looks like we went to some lengths to intentionally prevent importing the root container's triples.  

https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/importer/Importer.java#L360-L362

By removing that test and fixing up some other bits of code to accommodate it, the tests pass and I imagine the behavior is fixed for you.  I don't see any real problem with it, but I worry we might have had a good reason to intentionally prohibit importing the repository root, even if we didn't find a way to capture that reasoning in our integration tests.  The git blame doesn't help, as it implicates me for introducing that behavior and I don't remember why.  @escowles, do you recall any reasons why we wouldn't want to import the repository root?

escowles:
@mikedurbin I took a look at git-blame to see when we inserted that logic to stop importing the root node, and it looks like you are the one who did it:

https://github.com/fcrepo4-labs/fcrepo-import-export/blame/master/src/main/java/org/fcrepo/importexport/importer/Importer.java#L360

Looking at those tickets, I don't see anything about the root node specifically, just general import failures.  So if you don't remember anything about why we did that, I'm fine with removing it.

PR #111, opened 2017-09-03
Add missing pluginManagement tag

bbpennel:
Add missing pluginManagement tag to allow new version of fcrepo to download and to resolve m2e error

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,7 +106,7 @@
   </scm>
 
   <build>
-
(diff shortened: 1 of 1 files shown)


whikloj:
@bbpennel why move all the plugins to pluginManagement? My reading is that this is too allow other projects to inherit these plugin configurations, does anything inherit from the `fcrepo-import-export` tool?

PR #112, opened 2017-09-22
Updated bag generation to conform to APTrust specifications.

mikedurbin:
https://jira.duraspace.org/browse/FCREPO-2441

I tested these changes with a relatively large and complex export and the ingest into AP Trust has passed the validate phase without error.


awoods:
I am not sure what the issue is with Jenkins, but it builds locally.

PR #115, opened 2018-03-30
Add an explicit -repositoryRoot (-R) option.

peichman-umd:
Fixes https://jira.duraspace.org/browse/FCREPO-2741

### DIFF
--- a/src/main/java/org/fcrepo/importexport/ArgParser.java
+++ b/src/main/java/org/fcrepo/importexport/ArgParser.java
@@ -137,15 +137,15 @@ public class ArgParser {
                 .longOpt("writeConfig")
                 .hasArg(true).numberOfArgs(1).argName("writeConfig")
                 .desc("When present this flag indicates that a sample config should be written at the" +
-                       " specified filename.")
+                        " specified filename.")
                 .required(false).build());
 
         // Overwrite Tombstones
         configOptions.addOption(Option.builder("t")
-                 .longOpt("overwriteTombstones")
-                 .hasArg(false)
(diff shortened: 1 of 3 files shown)


src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -193,6 +193,13 @@
                 .desc("Path to the bag config file")
                 .build());
 
+        configOptions.addOption(Option.builder("R")
+                .longOpt("repositoryRoot").argName("uri")
```
bseeger: argName("uri") is called twice here.

peichman-umd: I was following the model of the other options, where `argName` is called for both the long opt and short option.

bseeger: Odd, it doesn't seem like calling it twice would do anything unique here. _But_ I can't say I fully understand what Option.Builder is doing in the first place.

src/main/java/org/fcrepo/importexport/ArgParser.java
```diff
@@ -193,6 +193,13 @@
...
+                .hasArg(true).numberOfArgs(1).argName("uri")
+                .required(false)
+                .desc("Repository root URI")
```
dbernstein: @peichman-umd :  Can you expand the description to include the intent of this feature, including whether or not it applies to both import and export modes as well as the default behavior when it is not specified?

peichman-umd: @dbernstein Opened new PR at #116

PR #122, opened 2019-03-25
Completes export of versions and external content handling.

dbernstein:
All unit tests pass.  Only ExporterIT is currently aligned.  Other integration tests have been
@ignored until they have been brought into alignment.

Resolves:  https://jira.duraspace.org/browse/FCREPO-2979

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.fcrepo</groupId>
     <artifactId>fcrepo-parent</artifactId>
-    <version>4.6.0</version>
+    <version>5.0.2</version>
   </parent>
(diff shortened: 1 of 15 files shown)


bseeger:
I'm currently looking through this. Builds fine, but doing a little manual testing.

bseeger:
I reviewed these changes and, code-wise, they look fine.  Things build just fine and an export functions against fcrepo4 code base (master branch).  I did notice that a proxied file is exported in whole - which seems like it should not be.  I think that's an issue that needs to be talked about more.

dbernstein:
@bseeger :  I have a feeling the export/import will be a work in progress.  I would say merge it if it seems okay for now and we can always create new tickets to address issues that are identified.

PR #126, opened 2019-07-03
Ensures that objects of rdf:type predicates in the ldp: namespace are…

dbernstein:
… removed before import.

This fix also brings the ImporterIT integration tests back into the mix (ie they
are no longer ignored).

Resolves: https://jira.duraspace.org/browse/FCREPO-3001
*********
This PR widens the ban on forbidden types on import to all types in the LDP namespace.  The functionality is covered by integration tests.  For manual test: 

1.  start a fresh repo
2. create a container resource1 and a nested binary
```
curl -X PUT -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/container
```
3. Add a triple
```
curl -X PATCH -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/container --data-binary "INSERT { <> <http://example.org/test> \"something\" } WHERE {}" 
```
4. Create a binary
```
curl -X PUT -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/container/binary --data-binary "hello world" -H "Content-Type: text/plain"
```
5.  Run the import export tool: 
```
java -jar ~/code/fcrepo-import-export/target/fcrepo-import-export-0.4.0-SNAPSHOT.jar  --mode export --predicates "http://www.w3.org/ns/ldp#contains" --resource "http://localhost:8080/rest" --versions --binaries --rdfLang "text/turtle" --dir export-dir
```
6. start a fresh repo:  stop report, remove fcrepo4-data and start the repo
7.  Import the data: 
```
java -jar ~/code/fcrepo-import-export/target/fcrepo-import-export-0.4.0-SNAPSHOT.jar  --mode import --predicates "http://www.w3.org/ns/ldp#contains" --resource "http://localhost:8080/rest" --versions --binaries --rdfLang "text/turtle" --dir export-dir
```
8.  Verify the resources have been successfully imported.

### DIFF
--- a/src/main/java/org/fcrepo/importexport/importer/Importer.java
+++ b/src/main/java/org/fcrepo/importexport/importer/Importer.java
@@ -36,6 +36,7 @@ import static org.fcrepo.importexport.common.FcrepoConstants.LAST_MODIFIED_BY;
 import static org.fcrepo.importexport.common.FcrepoConstants.LAST_MODIFIED_DATE;
 import static org.fcrepo.importexport.common.FcrepoConstants.LDP_NAMESPACE;
 import static org.fcrepo.importexport.common.FcrepoConstants.MEMBERSHIP_RESOURCE;
+import static org.fcrepo.importexport.common.FcrepoConstants.MEMENTO_NAMESPACE;
 import static org.fcrepo.importexport.common.FcrepoConstants.NON_RDF_SOURCE;
 import static org.fcrepo.importexport.common.FcrepoConstants.PAIRTREE;
 import static org.fcrepo.importexport.common.FcrepoConstants.RDF_TYPE;
@@ -574,6 +575,7 @@ public class Importer implements TransferProcess {
      */
     private boolean forbiddenType(final Resource resource) {
         return resource.getNameSpace().equals(REPOSITORY_NAMESPACE)
+            || resource.getNameSpace().equals(MEMENTO_NAMESPACE)
             || resource.getNameSpace().equals(LDP_NAMESPACE);
     }
 
@@ -604,10 +606,11 @@ public class Importer implements TransferProcess {
         ensureExists(parent(uri));
 
         final FcrepoResponse response;
+        final ByteArrayInputStream emptyStream = new ByteArrayInputStream(new byte[]{});
         if (fileForBinaryURI(uri, false).exists() || fileForBinaryURI(uri, true).exists()) {
-            response = client().put(uri).body(new ByteArrayInputStream(new byte[]{})).perform();
+            response = client().put(uri).body(emptyStream).perform();
         } else if (fileForContainerURI(uri).exists()) {
-            response = client().put(uri).body(new ByteArrayInputStream("".getBytes()), "text/turtle").perform();
+            response = client().put(uri).body(emptyStream, "text/turtle").perform();
         } else {
             return;
         }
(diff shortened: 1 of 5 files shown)


src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -610,8 +607,7 @@ private void makePlaceholder(final URI uri) throws IOException, FcrepoOperationF
         if (fileForBinaryURI(uri, false).exists() || fileForBinaryURI(uri, true).exists()) {
             response = client().put(uri).body(new ByteArrayInputStream(new byte[]{})).perform();
         } else if (fileForContainerURI(uri).exists()) {
-            response = client().put(uri).body(new ByteArrayInputStream(
-                    "<> a <http://www.w3.org/ns/ldp#Container> .".getBytes()), "text/turtle").perform();
+            response = client().put(uri).body(new ByteArrayInputStream("".getBytes()), "text/turtle").perform();
```
bbpennel: It seems like this is basically the same as the preceding block, except for the content type and the way an empty byte array is arrived at. It might make sense to merge the two, or create the body content in the same way in both cases?

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -574,10 +573,8 @@ private Model sanitize(final Model model) throws IOException, FcrepoOperationFai
      * @return true if the resource represents a type that may not be added/removed explicitly
      */
     private boolean forbiddenType(final Resource resource) {
-         return resource.getNameSpace().equals(REPOSITORY_NAMESPACE)
-             || resource.getURI().equals(CONTAINER.getURI())
-             || resource.getURI().equals(NON_RDF_SOURCE.getURI())
-             || resource.getURI().equals(RDF_SOURCE.getURI());
+        return resource.getNameSpace().equals(REPOSITORY_NAMESPACE)
```
bbpennel: https://wiki.duraspace.org/display/FEDORA5x/What+are+the+Server+Managed+Triples states that the memento namespace is also server managed, and importing a resource with `rdf:type <http://mementoweb.org/ns#SomeType>` produces the following error and results in a container being created without any of the exported properties present:
```
ERROR (Importer) Error while importing /Users/bbpennel/git/fcrepo-import-export/export-dir/rest/witht.ttl (409): The http://www.w3.org/1999/02/22-rdf-syntax-ns#type predicate may not take an object in the server managed namespaces (http://mementoweb.org/ns#).
```

Seems like it would make sense to add the third disallowed namespace in here at this point.

dbernstein: good catch.

src/main/java/org/fcrepo/importexport/importer/Importer.java
dbernstein: Sure - I think that makes sense.

dbernstein:
Regarding the issues raised with logging:  
Re 1:  there is another ticket for that one (at least there is now):  https://jira.duraspace.org/browse/FCREPO-3020
Re 2: I suppose we can make that first one a debug message since it is not an error and creates too much noise.
Re 3:  The fcr:versions endpoint is addressed in a soon-to-be posted PR for https://jira.duraspace.org/browse/FCREPO-2982

dbernstein:
@bbpennel :  Thanks for the feedback.  I've addressed the issues you raised in the above commit.

PR #137, opened 2020-02-11
Use Available BagIt Manifests During Import

mikejritter:
Resolves: https://jira.lyrasis.org/browse/FCREPO-3208

Updates Importer to use a preferred manifest based on all available in a BagIt Bag

------------------------------------------------
To test:
1. Start a clean fedora instance
2. Create a binary resource in the new repository
3. Create a `bag-config.yml` with the following properites
```
bag-info.txt:
  External-Description: Sample bag export for fcrepo-3208
```
4. Run the exporter using the `default` bag profile and a directory name `fcrepo-3208`
```
java -jar target/fcrepo-import-export-0.4.0-SNAPSHOT.jar --mode export --resource http://localhost:8080/fcrepo/rest --dir fcrepo-3208 --binaries --bag-profile default --bag-config bag-config.yml --user fedoraAdmin:secret3
```
5. Create a sha256 manifest for the newly exported bag
```
cd fcrepo-3208 
find data -type f -print0 | xargs -0 -I {} sha256sum {} > manifest-sha256.txt                                                                                                                                               
```
6. Start a fresh repository
7. Run the import tool, with debug logging enabled to see the checksums used on import
```
java -Dfcrepo.log.importexport=DEBUG -jar target/fcrepo-import-export-0.4.0-SNAPSHOT.jar --mode import --resource http://localhost:8080/fcrepo/rest --dir fcrepo-3208 --binaries --bag-profile default --bag-config bag-config.yml -u fedoraAdmin:secret3
```
8. Verify that the resource is in the repository and has a sha256 digest
```
curl -u fedoraAdmin:secret3 http://localhost:8080/fcrepo/rest/86feb268-750e-4cde-9dcb-25e8197be1b6/fcr:metadata
```


dbernstein:
@mikejritter :  it would be helpful if you could add a recipe for running a manual test of this functionality as well.  Like "1. fire up fedora 2. PUT resource A.  3. export bag ..."  with command line call for each step to facilitate testing.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -118,18 +121,15 @@
...
     private final List<URI> relatedResources = new ArrayList<>();
     private final List<URI> importedResources = new ArrayList<>();
     private URI repositoryRoot = null;
 
-    private Bag bag;
-
-    private MessageDigest sha1;
-
-    private Map<String, String> sha1FileMap;
+    private Map<String, String> bagItFileMap;
```
awoods: A comment describing what this Map is would be helpful.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -844,15 +833,71 @@ private File directoryForContainer(final URI uri) {
      * Verify the bag we are going to import
      *
      * @param bagDir root directory of the bag
-     * @return true if valid
+     * @return the {@link Bag} if valid
      */
-    public static boolean verifyBag(final File bagDir) {
+    public Bag verifyBag(final Path bagDir) {
```
awoods: This method can be `private`.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -844,15 +833,71 @@ private File directoryForContainer(final URI uri) {
...
+        }
+    }
+
+    /**
+     * Query a {@link Bag} for the highest ranking {@link Manifest} and use that in order to populate the
+     * {@code bagItFileMap} and {@code digestAlgorithm} for use when importing files
+     *
+     * @param bag The {@link Bag} to read from
+     */
+    public void configureBagItFileMap(final Bag bag) {
```
awoods: This method can be `private`.

src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -844,15 +833,71 @@ private File directoryForContainer(final URI uri) {
...
      */
-    public static boolean verifyBag(final File bagDir) {
+    public Bag verifyBag(final Path bagDir) {
         try {
-            final Bag bag = BagReader.read(bagDir);
-            BagVerifier.isValid(bag, true);
-            return true;
+            final BagReader bagReader = new BagReader();
+            final Bag bag = bagReader.read(bagDir);
+            // BagVerifier.isValid(bag, true);
```
awoods: Is there a reason for removing the `BagVerifier.isValid()` call?

mikejritter: Initially this branch being done on top of [FCREPO-3184](https://jira.lyrasis.org/browse/FCREPO-3184) which was handling Bag validation but I broke it off on its own. I ended up  using the `verifyBag` method to read the Bag anticipating that the other work will do the full verification. This PR does pull in an updated version of the `bagit-java` library so it should be easy enough to add the `BagVerifier` back in here.

mikejritter: @awoods One thing to note about adding in the `BagVerifier` is that we'll need to update some of the unit/integration tests. This is from the bag under `src/test/resources/sample/bag` having an invalid manifest (which had been done intentionally). This can be remedied, of course, but I wanted to mention it as it will have an impact on the diff.

PR #138, opened 2020-02-12
Verify bag profiles conform to bagit-profiles-spec

mikejritter:
Resolves: https://jira.lyrasis.org/browse/FCREPO-3204

* Adds validation to the BagProfile to test a profile after it has been loaded
* Updates profiles to be conformant with bagit-profiles-spec
  * changes for aptrust merged from [dart](https://github.com/APTrust/dart/blob/master/plugins/setup/aptrust/bagit_profiles.json)
* Tests all profiles under `src/main/resources/profiles`

---------------------------------
As this adds unit tests to `BagProfileTest` in order to test the bag profiles, testing for this can be done through maven:

```
mvn -Dtest=BagProfileTest test
```

or to run the three new tests individually:
```
mvn -Dtest=BagProfileTest#testAllProfilesPassValidation test
mvn -Dtest=BagProfileTest#testInvalidBagProfile test
mvn -Dtest=BagProfileTest#testInvalidBagProfileSerializationTypo test
```


src/main/java/org/fcrepo/importexport/common/BagProfile.java
```diff
@@ -367,4 +389,109 @@ public void validateConfig(final BagConfig config) {
...
+     * {@link BagProfile#getAllowedTagAlgorithms()}
+     *
+     * Tag-Files-Allowed
+     * If specified, the {@link BagProfile#getTagFilesRequired()} must be a subset of
+     * {@link BagProfile#getTagFilesAllowed()}. If not specified, all tags must match the '*' glob
+     *
+     * Accept-BagIt-Version
+     * At least one version is required
+     */
+    public void validateProfile() {
```
awoods: This method is only used in tests. Would it make more sense to move it into the `src/test` directory?
I am reluctant to add test code into functional classes.

..unless you are planning on invoking this method differently in the future.

src/main/java/org/fcrepo/importexport/common/BagProfile.java
```diff
@@ -367,4 +389,109 @@ public void validateConfig(final BagConfig config) {
...
+        final StringBuilder errors = new StringBuilder();
+
+        // Bag-Profile-Info
+        final List<String> expectedInfoFields = Arrays.asList("Source-Organization", "External-Description", "Version",
```
awoods: Maybe use the constants defined in `BagConfig.java`?

src/main/resources/profiles/default.json
```diff
@@ -1,6 +1,7 @@
 {
    "BagIt-Profile-Info":{
       "BagIt-Profile-Identifier":"http://fedora.info/bagprofile/default.json",
+      "BagIt-Profile-Version":"1.3.0",
```
awoods: Should the `Version` value below also be `1.3` instead of `1.2`?

src/main/resources/profiles/metaarchive.json
```diff
@@ -56,6 +57,9 @@
    ],
    "Allow-Fetch.txt":false,
    "Serialization":"optional",
+   "Accept-Serialization": [
+      "application/tar"
+   ],
```
awoods: Should the `Accept-BagIt-Version` below also include `1.0`?

mikejritter: Updating the `Accept-BagIt-Version` to include `1.0` makes sense to me. I'll double check the other profiles as well to make sure they accept `1.0`.

mikejritter:
@awoods I just pushed new commits which should resolve your comments. I was briefly considering refactoring `validateProfile` since it is now being done within the test class and likely doesn't need to throw an exception, but left it as is.

awoods:
This should be good to go... once the merge conflict has been resolved.

PR #143, opened 2020-05-06
Remove code used by bagit-support

mikejritter:
Resolves: https://jira.lyrasis.org/browse/FCREPO-3233

* Extracted BagIt classes to a new [bagit-support](https://jira.lyrasis.org/browse/FCREPO-3233) library
* Replaced BagIt classes with the bagit-support equivalents
* Update some of the initialization for BagItDigests to conform to bag profiles correctly when exporting

-------------------------

## Notes

The way the BagItDigest init was updated in the Exporter is to query first for the `Manifests-Required`, then `Manifests-Allowed`, then fallback to `sha1`. For `Manifests-Required` it makes sense to use each digest algorithm. I opted to do the same for `Manifests-Allowed` though I'm not 100% about that choice. We could try to alter this by picking a single algorithm to use, or maybe creating an issue to allow the user to specify what algorithm(s) to use (which would require additional validation against the profile, which shouldn't be a big deal). 

-------------------------

## Testing

1. Start a fresh fedora repository, e.g. with fcrepo4-docker `FEDORA_TAG=5.1.0 docker-compose up -d`
2. Create a binary
3. Create a bag-config.yml for export
```
bag-info.txt:
  External-Description: Sample bag export for fcrepo-3233
```
4. Run the exporter
```
java -jar target/fcrepo-import-export-0.4.0-SNAPSHOT.jar --mode export --resource http://localhost:8080/fcrepo/rest --dir fcrepo-3233 --binaries --bag-profile beyondtherepository --bag-config bag-config.yml --user fedoraAdmin:secret3
```
5. Restart the fedora repository, e.g. with fcrepo4-docker `docker-compose down; FEDORA_TAG=5.1.0 docker-compose up -d`
6. Run the importer
```
java -jar target/fcrepo-import-export-0.4.0-SNAPSHOT.jar --mode export --resource http://localhost:8080/fcrepo/rest --dir fcrepo-3233 --binaries --bag-profile beyondtherepository --bag-config bag-config.yml --user fedoraAdmin:secret3
```
7. Check that all binaries have been reimported


src/main/java/org/fcrepo/importexport/importer/Importer.java
```diff
@@ -180,9 +179,8 @@ public Importer(final Config config, final FcrepoClient.FcrepoClientBuilder clie
     private void loadBagProfile(final String bagProfile) {
         try {
             // load the bag profile
-            final URL url = this.getClass().getResource("/profiles/" + bagProfile + ".json");
-            final InputStream in = url == null ? new FileInputStream(bagProfile) : url.openStream();
-            final BagProfile profile = new BagProfile(in);
+            // todo: same as export for profile init
```
awoods, line 182: Is there another ticket for completing this `todo`?

mikejritter: Looking at this todo it looks like this still needs to be done for this PR. I updated the `BagProfile` constructor to use the built in profiles, so this should be migrated to use that method.

PR #150, opened 2020-09-21
Import ACLs

mikejritter:
Resolves: https://jira.lyrasis.org/browse/FCREPO-3333

Potentially resolves https://jira.lyrasis.org/browse/FCREPO-2983 as well (see notes)

* Update Importer to use correct URI when importing acls
* Add cli argument for importing/exporting acls
* Remove use of deprecated APIs
* Remove unnecessary stream usage

--------------------------------------

# Notes

* I only tested using acls as they are the majority of what is outlined under the Web Access Control documentation. I'm not sure if there are other types of WebAC which should be tested for 2983.
* There is only a long option for the cli argument, if needed a short option can be added as well
* I had to enable relaxed handling (`-Dfcrepo.properties.management=relaxed`) in order for certain metadata to be accepted, e.g. `fedora:created`. Before doing this I was receiving errors when trying to upload both `fcr:acl` and `fcr:metadata` resources. I'm not sure if it's worth it to create a ticket to have the import-export-tool also have a strict option or possibly add a note in the readme about it.

--------------------------------------

# Testing

1. Start a fresh fedora repository, e.g. with the one-click server `java -Dfcrepo.properties.management=relaxed -jar fcrepo-webapp-5.1.1-jetty-console.jar`
2. Create a binary in the repository
3. Create an acl for the binary
    ```
    @prefix acl: <http://www.w3.org/ns/auth/acl#>.

    <#authz> a acl:Authorization;
       acl:accessTo </rest/foo>;
       acl:agent "testuser";
       acl:mode acl:Read.
    ```
4. Add the acl to the binary: `curl -X PUT -H "Content-Type: text/turtle" --data-binary @acl.ttl -u fedoraAdmin:secret3 "http://192.168.99.100:8080/rest/binary/fcr:acl"`
5. Create a bag-config.yml for export
   ```
   bag-info.txt:
     External-Description: BagIt Bag
   ```
5. Run the exporter, specifying that acls should be exported
   ```
   java -jar target/fcrepo-import-export-0.4.0-SNAPSHOT.jar --mode export --resource http://localhost:8080/rest --dir fcrepo-3333  --binaries --acls --bag-profile beyondtherepository --bag-config bag-config.yml --bag-algorithms sha1 --user fedoraAdmin:secret3
   ```
6. Check the exported bag for the acls
7. Refresh the fedora repository
    a. Stop the server
    b. Clear the data if necessary (e.g. rm the `fcrepo4-data` created by the one-click or `docker-compose down` for docker)
    c. Restart the server
8. Run the importer
   ```
   java -jar target/fcrepo-import-export-0.4.0-SNAPSHOT.jar --mode import --resource http://localhost:8080/rest --dir fcrepo-3333  --binaries --acls --bag-profile beyondtherepository --bag-config bag-config.yml --bag-algorithms sha1 --user fedoraAdmin:secret3
   ```
9. Check that the acl exists for the binary: `curl -u fedoraAdmin:secret3 http://192.168.99.100:8080/rest/binary/fcr:acl`

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The standalone import/export utility can be run in either of two ways:
 The first time you run the utility with command-line arguments, a configuration file containing your provided arguments will be written to a file, the location of which will be displayed at the command line.
 
 ```sh
-$ java -jar fcrepo-import-export.jar --mode export --resource http://localhost:8080/rest --dir /tmp/test --binaries
+$ java -jar fcrepo-import-export.jar --mode export --resource http://localhost:8080/rest --dir /tmp/test --binaries --acls
 INFO 15:15:10.048 (ArgParser) Saved configuration to: /tmp/importexport.config
 INFO 15:15:10.091 (Exporter) Running exporter...
 ```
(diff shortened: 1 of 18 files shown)


awoods (changes requested):
This looks and works well. Thanks.

One minor point, I get an 'error' messaging that the root ACL is not exported.
The behavior of not exporting the default root ACL is probably ok, but the log message should likely be updated in this case to indicate something like:
```
info("The repository default root ACL is not being exported")
```
Additionally, if I PUT a new ACL as the root ACL, it is properly exported, which is good.

awoods (approved):

PR #157, opened 2020-10-27
Update version to 1.1.0-SNAPSHOT.

dbernstein:


awoods (changes requested):
Maybe 1.0.0?

dbernstein:
Do we want our release to be 1.0.0?  If so I need to redo the release as it currently going to be released as 0.4.0.  In which case we'll plan to release as 1.0.0 and we'll want our development version will be 1.1.0-SNAPSHOT.

awoods:
It seems that the import/export tool is at a 1.0 stage, no?

dbernstein:
I suppose it is.

dbernstein:
I will proceed accordingly.

awoods (approved):

PR #162, opened 2021-07-14
parallelized export; added resume file

pwinckles:
**Jira**: https://fedora-repository.atlassian.net/browse/FCREPO-3742

### What this does

1. Parallelizes the export code
2. Creates a new file that captures resource URIs that either failed to be processed or where waiting to be processed when the application was killed
3. Allows exports to use the file created in step 2 to resume a failed export
4. Adds a periodic progress report, every 10,000 resources, like the following: `INFO 11:42:44.067 (Exporter) Progress report: Exported 520 resources in PT19.126342S at 1600461 bytes/sec`

### How should this be tested

1. Setup a F5 instance with data
2. Run the exporter as follows: `java -jar fcrepo-import-export-1.1.0-SNAPSHOT.jar -b --dir my-5.1.1-export --user fedoraAdmin:fedoraAdmin --mode export --repositoryRoot http://localhost:8080/rest --resource http://localhost:8080/rest --binaries --versions`
3. Kill (ctrl+c) the utility while it's still exporting
4. See the `remaining_TIMESTAMP.log` file in the current directory containing the URIs of the resources that were waiting to be exported when the export was killed
5. Run the exporter again as follows: `java -jar fcrepo-import-export-1.1.0-SNAPSHOT.jar -b --dir my-5.1.1-export --user fedoraAdmin:fedoraAdmin --mode export --repositoryRoot http://localhost:8080/rest --resourcesFile remaining_TIMESTAMP.log --binaries --versions`
6. See the exporter pick up where it left off

Additionally, the number of threads the exporter used can be adjusted using the `-T` option.

### DIFF
--- a/src/main/java/org/fcrepo/importexport/common/Config.java
+++ b/src/main/java/org/fcrepo/importexport/common/Config.java
@@ -24,6 +24,7 @@ import org.slf4j.Logger;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
@@ -76,6 +77,9 @@ public class Config {
     private String rdfLanguage = DEFAULT_RDF_LANG;
     private String username;
     private String password;
+
+    private Integer threadCount;
+    private Path resourceFile;
 
     private boolean auditLog = false;
 
@@ -596,6 +600,12 @@ public class Config {
         final String predicates = String.join(",", this.getPredicates());
         map.put("predicates", predicates);
         map.put("auditLog", Boolean.toString(this.auditLog));
+        if (threadCount != null) {
+            map.put("threadCount", threadCount.toString());
+        }
+        if (resourceFile != null) {
+            map.put("resourceFile", resourceFile.toAbsolutePath().toString());
+        }
         return map;
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/fcrepo/importexport/common/Config.java
```diff
@@ -657,4 +667,38 @@ public boolean isIncludeAcls() {
...
+    }
+
+    /**
+     * Set the number of threads to use, must be at least 1
+     *
+     * @param threadCount the number of threads to use
+     */
+    public void setThreadCount(final Integer threadCount) {
+        if (threadCount == null) {
+            this.threadCount = threadCount;
```
whikloj, line 685: It seems like this should be 
```suggestion
            this.threadCount = 1;
```

pwinckles: It's actually okay for this to be null, and it allows the default value to be calculated here: https://github.com/pwinckles/fcrepo-import-export/blob/FCREPO-3742/src/main/java/org/fcrepo/importexport/exporter/Exporter.java#L792

I think the bug here is that if the thread count is less than 1, it should just ignore the value and use the default, which may or may not be 1. That is to say, I'll ditch the `Math.max()` call.

src/main/java/org/fcrepo/importexport/exporter/Exporter.java
```diff
@@ -713,42 +728,201 @@ void writeResponse(final URI uri, final InputStream in, final List<URI> describe
...
+        public synchronized void shutdown() {
+            if (!shutdown) {
+                shutdown = true;
+
+                try {
+                    executorService.shutdown();
+                    drainTasks();
+                    logger.info("Waiting for inflight tasks to complete...");
+                    if (!executorService.awaitTermination(5, TimeUnit.MINUTES)) {
+                        logger.warn("Failed to shutdown executor service cleanly after 1 minute of waiting");
```
whikloj, line 882: I think the message here should be updated to "5 minutes of waiting"

whikloj (approved):
This works as described and provides a file of failures to attempt to reprocess later.

PR #163, opened 2022-07-14
improve shutdown impl to work on java 17+

pwinckles:
# What this does

When I originally wrote the code that shutsdown the executor, I used reflection to get the info necessary to save the tasks that hadn't been processed. This will not work in Java 17+. This PR reworks the shutdown to no longer require reflection.

# How to test

Start an export and abort it while it's still running using ctrl+c. Then, verify the inflight tasks were logged.


dbernstein (approved):
The code looks good and builds locally.  There is a problem with the CI build but that  appears to be related to the fact that this project is still on Travis and needs to be migrated to github actions.  I'll create a ticket for that and once I've confirmed all is well locally I will merge.  Thank you @pwinckles !

